PR #52, opened 2017-12-27
Ignore missing directories when parsing projects

briot:
Those include the object directories, which libadalang does not use.

### DIFF
--- a/ada/extensions/analysis/c_api/unit_providers/body
+++ b/ada/extensions/analysis/c_api/unit_providers/body
@@ -58,7 +58,8 @@
       end if;
 
       begin
-         Load (Project.all, Create (+PF), Env);
+         Load (Project.all, Create (+PF), Env,
+               Report_Missing_Dirs => False);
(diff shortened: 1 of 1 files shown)


pmderodat:
Hello @briot 

Thank you for your contribution! While the change looks great to me, could you please also join a regression testcase that exhibits the change of behavior?

I think we don’t need anything fancy: just create a Python API testcase, for instance in `ada/testsuite/tests/python/missing_project_dir` that just tries to load a project with a missing object directory using `libadalang.UnitProvider.for_project`. If I understand correctly, this currently will yield some error/warnings on stdout/stderr while, thanks to your patch, this should be silent instead.

Thank you in advance!

pmderodat:
This is now implemented (#312)

briot:
> This is now implemented (#312)


That’s great, thanks a lot !

pmderodat:
Sure, thank you for reporting this in the first place!

PR #85, opened 2018-06-14
Nameres & API improvements + playground handle project files

raph-amiard:


ada/extensions/playground
```diff
@@ -0,0 +1,29 @@
...
+        self.parser.add_argument(
+            '-P', '--project', type=str, default='', help="GPR project file"
+        )
+
+    def process_args(self):
+        from tempfile import mkdtemp
+        if not self.args.project:
+            with open("{}/default.gpr".format(mkdtemp()), 'w') as f:
+                f.write("project Default is end Default;")
+                self.args.project = f.name
```
pmderodat: If I understand correctly, this will clutter /tmp without ever cleaning it up. Not cool!

PR #105, opened 2018-07-24
Nameres work

raph-amiard:


ada/language/ast.py
```diff
@@ -283,7 +283,7 @@ def generic_instantiations_internal(r=T.EnvRebindingsType):
                 head.singleton.concat(tail))
         )
 
-    @langkit_property()
+    @langkit_property(call_memoizable=True)
     def logic_val(from_node=T.AdaNode.entity, lvar=LogicVarType,
```
pmderodat: Given that this property reads a logic variable (which can be modified as a side effect), I think this annotation would deserve a bona fide comment explaining why it is safe. Could you please add it?

PR #152, opened 2018-10-05
Topic/nameres fixes

raph-amiard:


ada/extensions/src/libadalang-auto_provider.adb
```diff
@@ -47,7 +47,7 @@ package body Libadalang.Auto_Provider is
    is
       use Ada.Strings.Wide_Wide_Unbounded;
 
-      FQN  : constant Unbounded_Text_Type_Array := CU.P_Fully_Qualified_Name;
+      FQN  : constant Unbounded_Text_Type_Array := CU.P_Syntactic_Fqn;
```
Roldak: P_Syntactic_FQN would be more appropriate IMO

PR #154, opened 2018-10-11
Work on find_all_references

Roldak:


ada/extensions/nodes/ada_node/bodies
```diff
@@ -35,3 +35,108 @@ begin
...
+               for J in 1 .. Imported_Packages.Count loop
+                  Insert (Libadalang.Env_Hooks.Name_To_Symbols
+                    (Imported_Packages.Item (J)));
+               end loop;
+            end;
+         end if;
+      end loop;
+
+      --  Add all implicit references:
+      --   - If this unit is body, there is an implicit reference to its
```
pmderodat: Typo: “is a body”

ada/language/ast.py
```diff
@@ -6414,6 +6414,25 @@ class DefiningName(Name):
...
+            lambda c: c.filter(lambda n: Not(n.is_null))
+            .mapcat(lambda n: Self.find_all_refs_in(n))
+        ).concat(x.cast(BaseId).then(lambda i: If(
+            Self.name_is(i.sym) & x.xref.then(lambda ref: ref.node == Self),
+            x.singleton,
+            No(AdaNode.entity.array)
+        )))
+
+    @langkit_property(public=True, return_type=AdaNode.entity.array)
+    def find_all_references(units=AnalysisUnit.array):
```
pmderodat: Please document this public property.

ada/language/ast.py
```diff
@@ -6414,6 +6414,25 @@ class DefiningName(Name):
         doc="Returns this DefiningName's basic declaration"
     )
 
+    @langkit_property(public=False, return_type=AdaNode.entity.array)
+    def find_all_refs_in(x=AdaNode.entity):
```
pmderodat: Even though this properly is an implementation helper, describing how it’s supposed to be used and what is does would not hurt :-)

ada/extensions/nodes/ada_node/bodies
```diff
@@ -35,3 +35,147 @@ begin
...
+        (Symbols : in Libadalang.Env_Hooks.Symbol_Type_Array)
+         return Boolean is
+
+         Unit : Internal_Unit := Libadalang.Env_Hooks.Fetch_Unit
+           (Node.Unit.Context, Symbols, Node.Unit,
+            Unit_Specification, True, False);
+      begin
+         if Unit = Target or else
+              (Transitive and then
+               Does_Import_Target (Unit)) then
```
pmderodat: Coding style stikes again! Could you please put boolean operators on the line after and put the then on its own line? So something like:
```ada
if Unit = Target
   or else (Transitive and then Does_Import_Target (Unit))
then
```

pmderodat: Actually this statement could be simplified:
```ada
return (Unit = Target
        or else (Transitive and then Does_Import_Target (Unit));
```

ada/extensions/nodes/ada_node/bodies
```diff
@@ -35,3 +35,147 @@ begin
...
+
+      for I in 1 .. Prelude.Count loop
+         if Prelude.Item (I).Kind = Ada_With_Clause then
+            declare
+               Imported_Packages : Bare_Name_List :=
+                 Bare_With_Clause (Prelude.Item (I)).F_Packages;
+            begin
+               for J in 1 .. Imported_Packages.Count loop
+                  if Handle_Unit_Name (Libadalang.Env_Hooks.Name_To_Symbols
+                                         (Imported_Packages.Item (J))) then
```
pmderodat: Also `then` on its own line.

ada/extensions/nodes/ada_node/bodies
```diff
@@ -35,3 +35,147 @@ begin
...
+         return False;
+      exception
+         when Property_Error =>
+            return False;
+      end Handle_Unit_Name;
+
+      AST     : Bare_Compilation_Unit := Bare_Compilation_Unit (Root (From));
+      Prelude : Bare_Ada_Node_List;
+
+      Memoized_Position : Analysis_Unit_Maps.Cursor :=
```
pmderodat: I would suggest to rename this to “From_Cursor”. It’s more descriptive as to what it contains, and I don’t think “memoized” helps understanding its role (the notion of memoized position itself is debatable IMO!)

ada/language/ast.py
```diff
@@ -6412,6 +6412,32 @@ class DefiningName(Name):
...
+        """
+        Searches all references to this defining name in the given node.
```
pmderodat: So if I understand correctly, its looking for all refs in the given `x` subtree, right? (not just `x` as a node)

ada/testsuite/tests/properties/find_all_references/test.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+do_test(find_name("a.ads", "A"))
+do_test(find_name("a.ads", "X"))
+do_test(find_name("a.adb", "Y"))
+do_test(find_name("a.adb", "Get_X"))
+do_test(find_name("a.ads", "U"))
+do_test(find_name("a.ads", "Rec_Type"))
+do_test(find_name("b.adb", "X"))
+do_test(find_name("c.ads", "Foo"))
```
pmderodat: It’s hard to understand what’s going on from reading the expected output. What about instead something like:
```python
for unit, name in [('a.ads', 'A'), ('a.ads', 'X'), …]:
    print('All references to {} from {}:'.format(name, unit))
    for ref in find_name(unit, name).p_find_all_references(all_units):
        print('    {} ({}, {})'.format(ref.text, ref.unit, ref.sloc_range)
```

ada/language/ast.py
```diff
@@ -6434,7 +6434,7 @@ def find_all_references(units=AnalysisUnit.array):
         units.
         """
         return Self.filter_is_imported_by(units, True).mapcat(
-            lambda u: Self.top_level_decl(u).then(
+            lambda u: u.root.then(
```
pmderodat: I recommend enhancing your testcase to exercize this enhancement :)

PR #161, opened 2018-10-25
Work on parts navigation RA17-054

raph-amiard:


ada/language/ast.py
```diff
@@ -6605,6 +6657,30 @@ def find_all_references(units=AnalysisUnit.array):
             )
         )
 
+    @langkit_property()
+    def find_di(bd=BasicDecl.entity):
```
Roldak: find_di ? Please find a more explicit name.

PR #166, opened 2018-11-07
RB07-019: Fix crash in find all refs regarding  compilation unit lists.

Roldak:


ada/extensions/nodes/ada_node/bodies
```diff
@@ -108,7 +108,13 @@ function P_Filter_Is_Imported_By
 
       --  Add all explicit references by processing "with" clauses.
 
-      Prelude := AST.F_Prelude;
+      if Root_Node.Kind /= Ada_Compilation_Unit then
+         --  todo: handle list of compilation units.
```
raph-amiard: As discussed live: Not OK with this code. But it's not needed any more with the new changes I've made in ada_language_server. Tell me what you want to do with this PR!

Roldak: Okay let's drop this PR

Roldak: I will readjust this commit because it still fixes a `Constraint_Error` and put it the new PR

PR #167, opened 2018-11-08
Work on error recovery in name res.

Roldak:


ada/language/ast.py
```diff
@@ -36,6 +37,44 @@ def default_origin():
...
 
+def default_imprecise_fallback():
+    """
+    Helper to return an imprecise fallback dynamic param spec which defaults to
+    False.
+    """
+    return (imprecise_fallback, False)
+
+
+def _referenced_decl_internal_impl(subject, try_immediate):
```
raph-amiard: Plz don't do a macro, we try to avoid them as much as possible and here I don't see any reason that this cannot be a function ?

ada/language/ast.py
```diff
@@ -36,6 +37,44 @@ def default_origin():
...
+    )
+
+    try_compute_logic_val = Try(
+        compute_logic_val,
+        LogicValResult.new(success=False, value=No(AdaNode.entity))
+    )
+
+    precise_body = compute_logic_val.value.cast_or_raise(T.BasicDecl.entity)
+
+    imprecise_body = Let(lambda v=try_compute_logic_val: Let(
```
raph-amiard: With that code  you'll get the imprecise_body even the user didn't ask for the imprecise fallback

ada/language/ast.py
Roldak: Okay I'll put that in another property.

ada/language/ast.py
Roldak: I don't think so, the Let only appears in the "then" branch of the If expression. I checked the generated code and it is as I expect? I would agree with you if this was a `Var` since those are a syntactic sugar for a top-level `Let`, but it's not the case here.

raph-amiard: Ah, OMG that's true, that's it you know the DSL better than I do!

I'm not sure I like this style though. Very much like the other comment I made, that's abusing the implicit macro system we have in the DSL. First, most of the time an expression assigned to a variable that is not in a Var() is an error. Second, that won't transition well to the concrete syntax - the code will just be inlined down, so might as well do it already :)

PR #186, opened 2018-12-17
First version of nameres database

raph-amiard:


nameres_db/REQUIREMENTS.dev
```diff
@@ -0,0 +1,3 @@
+pony
```
Roldak: what about `funcy`?

PR #189, opened 2019-01-04
P506-012: Fix match_signature when there are rebindings.

Roldak:


ada/language/ast.py
```diff
@@ -4173,9 +4173,9 @@ class BasicSubpDecl(BasicDecl):
     def get_body_in_env(env=T.LexicalEnv):
         return env.get(Entity.name_symbol, LK.flat, categories=noprims).find(
             lambda ent:
-            ent.cast(T.BaseSubpBody)._.subp_spec
-            .match_signature(Entity.subp_decl_spec, True)
-        ).cast(T.BaseSubpBody)
+            ent.node.as_bare_entity.cast(T.BaseSubpBody)._.subp_spec
```
raph-amiard: This logic is a bit complex, and as such needs some comments !

PR #196, opened 2019-01-24
Make SubpRenamingDecl and NullSubpDecl resolve their corresponding decl

### LINKED ISSUE #195: SubpRenamingDecl not found as body
A `SubpRenamingDecl` can be the body of a `SubpDecl`, but this is not identified by libadalang. For example:

```ada
package Foo is
   procedure Bar;
end Foo;

package body Foo is
   procedure Baz is
   begin
      null;
   end Baz;

   procedure Bar renames Baz;
end Foo;
```

GPS correctly identifies the relationship between the declaration and the body and allows jumping between them.

A `decl_part` property should exist on `SubpRenamingDecl` and `body_part` on `SubpDecl` should find the renaming. `previous_part` and `next_part` should also be supported.

itoijala:
This is done by making them subclasses of `BaseSubpBody` instead of `ClassicSubpDecl`.
This makes them work the same as `ExprFunction`.

This makes conceptual sense, since they work the same way as a normal `SubpBody`: they can declare a subprogram or complete a previous declaration.

Fixes #195.

### DIFF
--- a/ada/language/ast.py
+++ b/ada/language/ast.py
@@ -4335,31 +4335,8 @@ class SubpDecl(ClassicSubpDecl):
     aspects = Field(type=T.AspectSpec)
 
 
-class NullSubpDecl(ClassicSubpDecl):
-    aspects = Field(type=T.AspectSpec)
-
-
 class AbstractSubpDecl(ClassicSubpDecl):
     aspects = Field(type=T.AspectSpec)
-
-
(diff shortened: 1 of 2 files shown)


raph-amiard:
Hello and thanks a lot for this pull request !

It makes a lot of conceptual sense indeed, and was exactly what I intended to do to fix #195 

I see that you took care that the testsuite passes, so for me that's an OK :)

PR #203, opened 2019-02-05
Fix a bunch of name resolution bugs. Some public API work

raph-amiard:


ada/language/ast.py
```diff
@@ -6066,7 +6061,10 @@ def direct_parent_aggregate():
     @langkit_property()
     def xref_equation():
         return If(
-            Self.parent.is_a(AspectClause, AspectAssoc),
+            Or(
```
Roldak: why not `Self.parent.is_a(AspectClause, AspectAssoc, PragmaArgumentAssoc)`?

raph-amiard: You're right, fixed!

PR #222, opened 2019-03-26
S326-012: Fix copy-paste bug, add test for expr_eval.

Roldak:


ada/testsuite/tests/ada_api/static_expr_eval/test.out
```diff
@@ -7,24 +7,24 @@ Expr <ObjectDecl ["Expr_2"] 8:4-8:26> evaluated to <Eval_Result ENUM_LIT <EnumLi
 Expr <ObjectDecl ["Expr_3"] 9:4-9:27> evaluated to <Eval_Result INT 12>
    Int value is 12
 
-Expr <ObjectDecl ["Expr_4"] 10:4-10:27> evaluated to <Eval_Result REAL  1.20000000000000E+01>
+Expr <ObjectDecl ["Expr_4"] 10:4-10:27> evaluated to  1.200E+01
```
pmderodat: If `Float_Precision` is supposed to be 2, why do we have three digits after the dot, here? ;-)

Roldak: because the exponent is 1. So the value is actually `12.00`, meaning we indeed have two digits of precision.

PR #225, opened 2019-03-28
More nameres

raph-amiard:


pmderodat:
Looks like you fixed a lot of bugs! Would it be possible to add regression tests for these?

PR #231, opened 2019-04-08
Nameres stuff

Roldak:


ada/language/ast.py
```diff
@@ -9151,10 +9156,14 @@ def array_attr_equation():
             # Prefix is not a type: In that case we have permission to resolve
             # prefix separately.
             Let(lambda
-                res=Entity.prefix.resolve_names_internal(True, LogicTrue()),
+                res=Entity.prefix.resolve_names_internal(
+                    True,
+                    Predicate(BaseTypeDecl.is_array_def_with_deref,
```
raph-amiard: Adding a comment there would be a good idea given the complexity of the logic here :)

PR #235, opened 2019-04-11
More nameres fixes!

Roldak:


ada/language/ast.py
```diff
@@ -2956,6 +2956,12 @@ def matching_assign_type(expected_type=T.BaseTypeDecl.entity):
                 .accessed_type.matching_assign_type(expected_type)
             ),
 
+            And(
```
raph-amiard: You should add comment about symmetry here

raph-amiard:
Argh just saw that you have regressions on the travis due to checking stuff in the runtime, which is . Let's discuss about that tomorrow, but we should use Test_Statement_UID in the failing test

PR #245, opened 2019-04-25
Nameres of record representation clauses & fix get_aspect

Roldak:


ada/language/ast.py
```diff
@@ -5025,6 +5025,25 @@ def xref_equation():
             LogicTrue()
         )
 
+    @langkit_property(return_type=T.String)
+    def aspect_name(n=T.Name.entity):
+        """
+        Return the string representation of the given name, which must be a
+        Name that can appear in an aspect association id.
+        """
+        # TODO: would be cleaner to implement a general "image" function in
```
raph-amiard: Indeed it would ಠ_ಠ

PR #248, opened 2019-04-26
Fix nameres issue on generics & fix referenced decl for generic subp

Roldak:


ada/language/ast.py
```diff
@@ -356,7 +356,18 @@ def referenced_decl_internal_helper(ref_var=T.LogicVar,
...
                 .value.cast_or_raise(T.BasicDecl.entity)
-        )
+        ).then(lambda x: x.match(
+            # If the logic variable is bound to a GenericSubpInternal, retrieve
+            # the instantiation leading to it instead.
+            lambda g=T.GenericSubpInternal: entity_no_md(
+                type=T.BasicDecl,
+                node=g.info.rebindings.new_env.env_node
+                .cast_or_raise(T.GenericInstantiation),
+                rebindings=x.info.rebindings.get_parent,
```
raph-amiard: A comment here would be helpful, explaining you're removing the instantiation from the list of rebindings.

Roldak: ok

PR #258, opened 2019-05-13
S327-012: prototypes for is_dispatching_call and is_static_call

Roldak:


ada/testsuite/tests/properties/is_dispatching_call/pkg.adb
```diff
@@ -0,0 +1,29 @@
+package body Pkg is
+   procedure Test is
+      type F_A_Type is access procedure (Self : T);
+
+      X : T'Class;
+      Y : T;
+      I : Int;
+      F_A : F_A_Type;
+   begin
+      -- normal dispatching proc
```
raph-amiard: Following live discussions:

1. Please check that your code compiles with GNAT
2. Please add the correct example of dispatching  on return type
3. Personal taste, but if possible, I prefer when test cases fit in one file, makes them much easier to browse.

ada/language/ast.py
```diff
@@ -6919,6 +6951,100 @@ def is_write_reference():
...
         )
 
+    @langkit_property(dynamic_vars=[default_imprecise_fallback()])
+    def potential_actuals_for_dispatch():
+        spec = Var(Entity.referenced_decl.subp_spec_or_null)
+
+        # Handle the case where the dispatch is done on the tag of the
+        # LHS expr of an assign statement. In that case, its expected type
+        # is the return type of the called subprogram.
+        # TODO: The assign statement might not be the direct parent, e.g
```
raph-amiard: So:

1. You're already doing better than GNAT here (as checked live)
2. If you add a qualified expression around the call, the target type will be statically known.
3. With a type conversion, it's no longer a controlling result call (apparently)

So you can proudly remove the TODO ;)

PR #259, opened 2019-05-15
Fix is_call returning true on keyword argument, fix navigation for protected type decls, fix is_float_type and is_fixed_point

Roldak:


ada/language/ast.py
```diff
@@ -10512,9 +10522,8 @@ class ProtectedBody(Body):
         dest_env=env.bind(
             Self.initial_env,
             # If this is a sub package, sub_package
-            Entity.body_decl_scope
-
-            ._or(Entity.body_scope(False))
+            Entity.body_scope(False, True)
```
raph-amiard: Wat is that ?

Roldak: The impl of body_decl_scope is a call to body_scope with arguments (True, True), meaning we "follow private". So, the "__next_part" symbol may end up in the private scope of the protected type decl after env spec is executed. To handle this case, we would have to reimplement next_part_for_decl specifically in a ProtectedTypeDecl to fetch "__next_part" in the private scope if it exists. 
If we just don't do that, the mapping will only ever exist in the public scope, and so we can simply result basic_decl_next_part_for_decl which I factored out above.

raph-amiard: I wonder if we should qualify the argument's name for all the calls to `body_scope`

PR #262, opened 2019-05-22
Fix is_call on generic subprograms and zip_with_params losing entity info, + nameres.

Roldak:


ada/language/ast.py
```diff
@@ -359,14 +359,19 @@ def referenced_decl_internal_helper(ref_var=T.LogicVar,
...
-                rebindings=x.info.rebindings.get_parent,
-                from_rebound=x.info.from_rebound
+                info=T.entity_info.new(
+                    # Since we return the instantiation itself, remove it from
+                    # its rebindings.
+                    rebindings=x.info.rebindings.get_parent,
+                    from_rebound=x.info.from_rebound,
+                    md=new_metadata(
+                        is_call=x.info.md.is_call,
+                        dottable_subp=x.info.md.dottable_subp
```
raph-amiard: I don't know if propagating dottable_subp is really relevant here, since a generic subprogram instantiation will never be dottable AFAICT.

ada/language/ast.py
```diff
@@ -7593,8 +7593,15 @@ def zip_with_params():
         return params.then(
             lambda _: Self.match_formals(params, Entity, is_dottable_subp).map(
                 lambda m:
-                ParamActual.new(param=m.formal.name.as_bare_entity,
-                                actual=m.actual.assoc.expr),
+                ParamActual.new(
```
raph-amiard: While that certainly works, I wonder if changing  SingleFormal.name to store an entity wouldn't be a cleaner fix. What do you think ?

ada/language/ast.py
```diff
@@ -6062,6 +6062,25 @@ def matching_nodes():
         """
         return env.bind(Self.node_env, Entity.env_elements)
 
+    @langkit_property(return_type=Equation, dynamic_vars=[env, origin])
+    def call_argument_equation(formal_type=T.BaseTypeDecl.entity,
+                               call_is_primitive_of=T.BaseTypeDecl):
+        """
+        Generate the equation that binds the type_var of the this expression
```
raph-amiard: "the this expression" ?

ada/language/ast.py
```diff
@@ -8738,6 +8738,25 @@ def dottable_subp_of():
             No(T.BaseTypeDecl.entity.array)
         ))
 
+    @langkit_property(return_type=T.EnvRebindings)
+    def shed_first_rebindings(rebindings=T.EnvRebindings,
+                              to_shed=T.EnvRebindings):
+        """
+        Remove ``to_shed`` from ``rebindings`` if those appear first in
+        ``rebidings``.
```
raph-amiard: typo "rebidings"

ada/testsuite/tests/name_resolution/primitive_return_type/test.adb
```diff
@@ -0,0 +1,18 @@
+procedure Main is
```
raph-amiard: Your code seems to suggest that this should also work in nested instantiations (eg. shed just the right amount of rebindings), however your test only contains one level of instantiation/nesting. Shouldn't you test with at least two ?

ada/testsuite/tests/name_resolution/primitive_return_type/test.adb
```diff
@@ -0,0 +1,18 @@
...
+   generic
+      type Dont_Care is private;
```
raph-amiard: Also what is this doing here ? Generics can have no params

raph-amiard:
SHAME ON YOU !!! A STYLECHECK IS FAILING

PR #268, opened 2019-06-02
Ada get_doc implem & nameres work

raph-amiard:


contrib/laldoc/laldoc/__init__.py
```diff
@@ -453,7 +363,7 @@ def append_decl(decl):
         decls = package_decl.f_public_part.f_decls
         types = {}
 
-        for decl in list(decls):
+        for decl in list(d for d in decls if d.is_a(lal.BasicDecl)):
```
pmderodat: This is a violation of our coding style:
> Avoid using comprehensions as the iterated value in a regular for loop.

We can replace the above with a simple conditioned `continue` in the loop body:
```python
if not decl.is_a(lal.BasicDecl):
    continue
```

ada/language/ast.py
```diff
@@ -1008,13 +1008,49 @@ def child_unit(name_expr, scope_expr, dest_env=None,
...
 class BasicDecl(AdaNode):
     """
     Root class for an Ada declaration (RM 3.1). A declaration associates a name
     with a language entity, for example a type or a variable.
     """
 
+    @langkit_property(public=True, external=True,
+                      return_type=DocAnnotation.array,
+                      uses_entity_info=False, uses_envs=False)
+    def get_doc_annotations():
```
pmderodat: For the public API, I would suggest to get rid of the `get_` prefix for functions, looks redundant (all functions return something, so they could all start with `get_`).

Also what about getting rid of the doc/documentation inconsistency with the other added property? All other names seem to use the abbreviated, so let’s use only this?

ada/extensions/nodes/basic_decl/bodies
```diff
@@ -59,3 +59,58 @@ begin
...
+           Create_Character_Type_Array (Key'Length);
+         DSL_Val : Character_Type_Array_Access :=
+           Create_Character_Type_Array (Val'Length);
+         Ann : Internal_Doc_Annotation;
+      begin
+         DSL_Key.Items := Key;
+         DSL_Val.Items := Val;
+         Ann.Key := DSL_Key;
+         Ann.Val := DSL_Val;
+         Ret.Items (Idx) := Ann;
```
pmderodat: I think this lacks an increment for `Idx`

ada/extensions/nodes/basic_decl/bodies
```diff
@@ -59,3 +59,58 @@ begin
...
+begin
+   for El in Doc.Annotations.Iterate loop
+      declare
+         Key : Text_Type := Annotations_Maps.Key (El);
+         Val : Text_Type := Annotations_Maps.Element (El);
+         DSL_Key : Character_Type_Array_Access :=
+           Create_Character_Type_Array (Key'Length);
+         DSL_Val : Character_Type_Array_Access :=
+           Create_Character_Type_Array (Val'Length);
+         Ann : Internal_Doc_Annotation;
```
pmderodat: Misaligned colons (and `constant` keywords)

ada/extensions/src/libadalang-doc_utils.adb
```diff
@@ -0,0 +1,150 @@
...
+
+package body Libadalang.Doc_Utils is
+
+   use XStrings;
+
+   ----------------------
+   -- Extract_Doc_From --
+   ----------------------
+
+   function Extract_Doc_From
```
pmderodat: Please add a forward declaration + documentation for this function

ada/extensions/src/libadalang-doc_utils.adb
```diff
@@ -0,0 +1,150 @@
...
+      while Tok /= No_Token  loop
+         K := Kind (Data (Tok));
+         case K is
+            when Ada_Whitespace =>
+               T := To_XString (Text (Tok));
+               exit when T.Count (LF) > 1;
+
+            when Ada_Comment =>
+               T := To_XString (Text (Tok));
+               T := T.Slice (3, T.Length);
```
pmderodat: Please propagate the `Strip the "--"` comment from the previous implementation

ada/extensions/src/libadalang-doc_utils.adb
```diff
@@ -0,0 +1,150 @@
...
+
+               --  If this is an annotation then
+               if T.Starts_With ("%")  then
```
pmderodat: Double space before `then`

ada/extensions/src/libadalang-doc_utils.adb
```diff
@@ -0,0 +1,150 @@
...
+                  declare
+                     --  Try to split on the ":"
+                     X : XString_Array := T.Split (":");
+                     K : Wide_Wide_String :=
+                       X (1).Slice (3, X (1).Length).To_String;
+                     --      ^ Strip % prefix
```
pmderodat: This is incomplete: unless we strip/trim `X (1).Slice (2, X (1).Length)` (which is what the previous implem did), we must check that `X (1).Element (2)` is a space

ada/extensions/src/libadalang-doc_utils.adb
```diff
@@ -0,0 +1,150 @@
...
+
+      --  Reverse the Doc vector if lines were searched backwards
+
+      if Backwards then
+         Doc_Vec.Reverse_Elements;
+      end if;
+
+      --  Transform the doc vector into a string
+      declare
+         Last_Index : Natural := Doc_Vec.Last_Index;
```
pmderodat: Missing `constant` keyword

ada/extensions/src/libadalang-doc_utils.adb
```diff
@@ -0,0 +1,150 @@
...
+      begin
+         for I in Doc_Vec.First_Index .. Doc_Vec.Last_Index loop
+            declare
+               L : XString renames Doc_Vec (I);
+            begin
+               Ret.Doc.Append (L.Slice (3, L.Length));
```
pmderodat: I don’t understand the slicing here: the `--` prefix was removed before appending the string to `Doc_Vec`, so not sure if this is correct/necessary. Can you adjust or add a comment?

Also, what about the removal of common indentation? (was present in previous implementation)

ada/extensions/src/libadalang-doc_utils.adb
```diff
@@ -0,0 +1,150 @@
...
+   begin
+      if Decl.Kind in Ada_Base_Package_Decl
+        and then Decl.Kind /= Ada_Generic_Package_Decl
+      then
+         return Extract_Doc_From
+           (Decl.Token_Start, Backwards => True, Skip_White_Lines => -1);
+      else
+         return Extract_Doc_From
+           (Decl.Token_End, Backwards => False, Skip_White_Lines => 1);
+      end if;
```
pmderodat: Please comment this logic (there were useful comments in the previous implementation). Also, the previous implementation seemt to handle more cases: documenation before the `package` keyword, but also documentation before the unit prelude for library-level units (which we do have in LAL). Have you checked the result on the generated documentation?

ada/language/ast.py
```diff
@@ -1008,13 +1008,49 @@ def child_unit(name_expr, scope_expr, dest_env=None,
...
+        """
+        Return the documentation annotations associated with this decl.
+        Annotations are any comment line of the form::
+
+            --% [annotation_name]: [annotation]
+
+        ALPHA! NOTE: For the moment, this function uses very simple hard-coded
```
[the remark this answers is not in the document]
raph-amiard: SGTM!

ada/extensions/src/libadalang-doc_utils.adb
raph-amiard: Correct, thank you for the catch

raph-amiard:
@pmderodat fixed all your comments and added a test!

ada/extensions/nodes/basic_decl/bodies
```diff
@@ -59,3 +59,58 @@ begin
...
+
+function P_Get_Documentation
+  (Node : access Bare_Basic_Decl_Type'Class) return Character_Type_Array_Access
+is
+   use Libadalang.Doc_Utils;
+   Decl : constant Libadalang.Analysis.Basic_Decl
+     := Wrap_Node (Node).As_Basic_Decl;
+   Doc : Doc_Type := Get_Documentation (Decl);
+   Ret : Character_Type_Array_Access
+     := Create_Character_Type_Array (Doc.Doc.Length);
```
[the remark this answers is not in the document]
pmderodat: Thanks! What about `constant` qualifiers?

ada/extensions/src/libadalang-doc_utils.adb
```diff
@@ -136,12 +136,29 @@ package body Libadalang.Doc_Utils is
...
             declare
                L : XString renames Doc_Vec (I);
             begin
-               Ret.Doc.Append (L.Slice (3, L.Length));
+
+               --  Check that every character we're going to strip is a white
+               --  space; else, raise an error.
+               if not (for all C of L.Slice (1, Offset - 1) => Is_Space (C))
+               then
+                  raise Property_Error with "Invalid formatted docstring";
```
pmderodat: I believe “docstring” designates string literals (and will thus be surprising for a lot of Ada programmers): here we’re just processing comments, so I think we should reword to something like “Invalid indentation in documentation comment”.

ada/extensions/src/libadalang-doc_utils.adb
```diff
@@ -136,12 +136,29 @@ package body Libadalang.Doc_Utils is
...
       declare
          Last_Index : Natural := Doc_Vec.Last_Index;
+
+         First_Line : XString :=
+           (if Last_Index /= 0 then Doc_Vec.First_Element else Null_XString);
+
+         Offset     : Positive :=
+           First_Line.Length - First_Line.Trim.Length + 1;
+         --  Offset for the first non whitespace char on the first line. Will
+         --  be used as the offset to start every subsequent doc line.
```
pmderodat: The previous code was taking the minimum indentation (and thus never failed) while here we take the indentation of the first line and raise an error if it’s not the minimum indentation. Given that this property is just for laldoc’s purposes, i.e. for doc that we write ourselves, I guess that’s fine, but I hope we won’t get libadalang-doc build failures because of poorly formatted comments in the future (the testsuite does not catches that for now).

ada/extensions/src/libadalang-doc_utils.adb
```diff
@@ -30,6 +30,15 @@ package body Libadalang.Doc_Utils is
 
    use XStrings;
 
+   function Extract_Doc_From
+     (Token            : Token_Reference;
+      Backwards        : Boolean;
+      Skip_White_Lines : Integer := -1) return Doc_Type;
+   --  Extract documentation from ``Token``. If ``Backwards`` is ``True``,
+   --  then search for documentation backwards. Skip up to ``Skip_White_Lines``
+   --  white lines separating the first doc comment from ``Start_Token``. If
```
pmderodat: `Start_Token` above should be `Token`

ada/language/ast.py
```diff
@@ -8751,12 +8773,8 @@ def return_type():
                     Entity.info.from_rebound
                 ),
 
-                entity_no_md(
-                    BaseTypeDecl,
-                    Entity.info.md.primitive_real_type.cast(BaseTypeDecl),
-                    Entity.info.rebindings,
-                    Entity.info.from_rebound
-                ),
```
Roldak: Would be nice to have a small comment explaining why this new logic is the right one

ada/extensions/src/libadalang-doc_utils.adb
raph-amiard: For the moment, the only use case of get_doc is to generate sphinx. With the old algorithm, you would get very weirdly formatted sphinx anyway if you used the old algorithm in a case this algorithm doesn't handle, and probably not what you wanted in the first place (like starting with an indented block)

raph-amiard: And actually the testsuite does have a case of unproperly formatted comment so I'm not sure what you mean ?

ada/extensions/src/libadalang-doc_utils.adb
raph-amiard: Right, good catch !

ada/extensions/src/libadalang-doc_utils.adb
raph-amiard: Good point

ada/language/ast.py
raph-amiard: What if I have no idea ?

Roldak: We're good to go then :+1:

PR #275, opened 2019-06-10
PrimTypeAccessor.prim_type: make this field private

pmderodat:


pmderodat:
Thanks! Good point, not sure about how to do it exactly: private fields in struct probably don’t make sense, so in the end maybe we should just remove the privacy setting on UserFields altoghether and use whatever is appropriate.

PR #278, opened 2019-06-24
Rework is_call property and fix instantiation of a formal package with a renaming one.

Roldak:


ada/language/ast.py
```diff
@@ -407,7 +385,7 @@ def logic_val(from_node=T.AdaNode.entity, lvar=LogicVar,
         success = Var(If(
             try_immediate & Not(lvar.get_value.is_null),
             True,
-            from_node.parents.find(lambda p: p.xref_entry_point).resolve_names
+            from_node.resolve_names_from_closest_entry_point
```
raph-amiard: Since you're at that, p_xref_entry_point doesn't follow RM terminology.

The ARM calls what we call an xref_entry_point a "complete context for name resolution":

> 4   Overload resolution is applied separately to each complete context, not
> counting inner complete contexts. Each of the following constructs is a
> complete context:
> 
> 5     * A context_item.
> 
> 6     * A declarative_item or declaration.
> 
> 7     * A statement.
> 
> 8     * A pragma_argument_association.
> 
> 9     * The expression of a case_statement.

I wonder if we shouldn't change terminology to fit the RM here

ada/language/ast.py
```diff
@@ -7586,7 +7586,7 @@ def check_for_type(typ=T.BaseTypeDecl.entity):
 
             Entity.parent.cast(T.CallExpr).then(
                 lambda ce: ce.check_for_type(
-                    bind_origin(Self, typ.expr_type)
+                    bind_origin(Self, typ.comp_type(True))
```
raph-amiard: plz use named notation for bool params. Also, all the rest of this property seems comment on at least which use case we are in, and this is not the case here for the default case. Could you also add a comment ? kthx

ada/language/ast.py
```diff
@@ -407,7 +407,7 @@ def logic_val(from_node=T.AdaNode.entity, lvar=LogicVar,
         success = Var(If(
```
raph-amiard: English nit about your commit message: You'd say "Rework is_call completely" rt. "Rework completely is_call". Nobody cares though. Just sayin'

ada/language/ast.py
```diff
@@ -6714,7 +6717,10 @@ def parent_name_equation(typ=T.BaseTypeDecl.entity, root=T.Name):
                 lambda ce=T.CallExpr:
                 ce.as_entity.subscriptable_type_equation(typ),
 
-                lambda ed=T.ExplicitDeref: ed.as_entity.eq_for_type(typ),
+                lambda ed=T.ExplicitDeref:
```
raph-amiard: Not clear what this is doing in this commit. Isn't that a fix in itself ? also not clear because this change is not commented, neither here or in the commit msg.

EDIT: Ok I understand now. Please comment what you are doing here (eg. storing a ref to the subp spec for later use in `is_call`, if I understand correctly)

ada/language/ast.py
```diff
@@ -7521,6 +7567,8 @@ def subprogram_equation(subp_spec=T.BaseFormalParamHolder.entity,
             # subprogram.
             TypeBind(Self.type_var, subp_spec.return_type)
 
+            & Bind(Self.subp_spec_var, subp_spec)
```
raph-amiard: Same, add comment

ada/language/ast.py
```diff
@@ -7765,10 +7818,10 @@ def general_xref_equation(root=(T.Name, No(T.Name))):
                 & Bind(Self.ref_var, Self.prefix.ref_var)
                 & Entity.eq_for_type(typ)
                 & typ.access_def.cast(AccessToSubpDef).then(
-                    lambda _: Self.parent_name(root).as_entity.then(
+                    lambda ad: Self.parent_name(root).as_entity.then(
                         lambda pn: pn.parent_name_equation(typ, root),
                         default_val=LogicTrue()
-                    ),
+                    ) & Bind(Self.subp_spec_var, ad.subp_spec),
```
raph-amiard: Same

ada/language/ast.py
```diff
@@ -7200,7 +7211,7 @@ def is_dispatching_call():
         .. note:: This is an experimental feature. There might be some
             discrepancy with the GNAT concept of "dispatching call".
         """
-        return Entity.is_call & Let(
+        return Entity.is_direct_call & Let(
```
raph-amiard: 1. your commit does more than what it says it does. Apparently you're also "fixing" is_dispatching/static_call.

2. I wonder if this is correct. A call to a subprogram access is a dispatching call by definition, so I wonder if it shouldn't be:

```python
Not(Entity.is_direct_call) | Let (
...
```

ada/language/ast.py
```diff
@@ -7236,7 +7247,7 @@ def is_static_call():
         .. note:: This is an experimental feature. There might be some
             discrepancy with the GNAT concept of "static call".
         """
-        return Entity.is_call & Not(Entity.is_dispatching_call)
+        return Entity.is_direct_call & Not(Entity.is_dispatching_call)
```
raph-amiard: If what I said above is correct, this should make this redundant. Also please put those changes in a subsequent commit :P

ada/language/ast.py
```diff
@@ -5704,14 +5704,32 @@ class PackageRenamingDecl(BasicDecl):
     renames = Field(type=RenamingClause)
     aspects = Field(type=T.AspectSpec)
 
+    @langkit_property(return_type=T.BasicDecl.entity)
+    def renamed_package():
```
raph-amiard: This might be a good candidate for `public=True` :)

ada/language/ast.py
```diff
@@ -5704,14 +5704,32 @@ class PackageRenamingDecl(BasicDecl):
...
+        """
+        Return the declaration of the package that is renamed by Self.
+        """
+        return env.bind(
+            Entity.node_env,
+            Entity.renames.renamed_object.env_elements.at(0)._.cast(BasicDecl)
+        )
+
+    @langkit_property(return_type=T.BasicDecl.entity)
+    def final_renamed_package():
```
raph-amiard: Same comment

ada/language/ast.py
```diff
@@ -8212,6 +8230,11 @@ def pkg_env(pkg=T.BasicDecl.entity):
         visibility on the formals.
         """
 
+        pkg = Var(from_pkg.cast(PackageRenamingDecl).then(
```
raph-amiard: This logic needs to be commented, because it's extremely complex !

ada/testsuite/tests/name_resolution/formal_pkg_as_renaming_pkg/d_data.ads
```diff
@@ -0,0 +1,6 @@
+with Data_Pkg;
+
+package D_Data is
+   package Pkg_H is new Data_Pkg (Value_T => Integer);
+   package Pkg renames Pkg_H;
```
raph-amiard: Since you implemented resolution of several levels of renamings, should be good to test with several levels of renamings.

ada/language/ast.py
```diff
@@ -6736,8 +6736,15 @@ def parent_name_equation(typ=T.BaseTypeDecl.entity, root=T.Name):
                 lambda _: Bind(Self.type_var, No(T.AdaNode.entity).node),
             ) & Self.parent_name(root).as_entity.then(
                 lambda pn: pn.parent_name_equation(
-                    typ.comp_type(
-                        is_subscript=Not(Self.is_a(T.ExplicitDeref))
+                    # An ExplicitDeref will access the comp_type itself, so
```
raph-amiard: This comment seems to apply to the `typ.comp_type` below, not to the bit of special logic you added for access to subps. please:

1. Move the comment in the correct location
2. Add a comment about the new bit of logic

ada/language/ast.py
Roldak: I think the best one would be "Completely rework is_call" :P

raph-amiard: is_call, completely. Rework!

ada/testsuite/tests/name_resolution/formal_pkg_as_renaming_pkg/d_data.ads
```diff
@@ -2,5 +2,10 @@ with Data_Pkg;
 
 package D_Data is
    package Pkg_H is new Data_Pkg (Value_T => Integer);
-   package Pkg renames Pkg_H;
+   package Pkg_1 renames Pkg_H
+   package Pkg_2 renames Pkg_1
+   package Pkg_3 renames Pkg_2
+   package Pkg_4 renames Pkg_3
+   package Pkg_5 renames Pkg_4
+   package Pkg renames Pkg_5;
```
Roldak: @raph-amiard Is that enough renamings? :P

raph-amiard: @Roldak no, please add more

PR #288, opened 2019-07-10
P506-012: Fix UnconstrainedArrayIndices.index_type

raph-amiard:


pmderodat:
Looks like you forgot the fix for S709-005: the new testcase fails on Travis CI. Could you have a look?

raph-amiard:
@pmderodat I think it's only because my branches were not named the same, and the fix is in Langkit. Restarting the travis !

raph-amiard:
Actually the travis is passing now, didn't have to restart it :)

pmderodat:
I see, thanks!

pmderodat:
The Travis “passes” (it’s green), but if you have a look at the testsuite log there, you’ll see one failure :). I see in the log that there is no `topic/fix_index_type` branch in your Langkit fork, so that explains it.

PR #290, opened 2019-07-24
S724-021: Improve binding to for_project

danielmercier:
Pass the runtime, target and scenario variables to the function
for_project, to be able to correctly call create_project_unit_provider.


ada/testsuite/tests/ocaml_api/project_unit_provider/p.gpr
```diff
@@ -1,3 +1,5 @@
 project P is
-   for Source_Dirs use ("src");
+   type Src_Dir_Type is ("src1", "src2");
+   Src_Dir : Src_Dir_Type := external ("SRC_DIR");
+   for Source_Dirs use (Src_Dir);
```
pmderodat: Looks like you removed the source files (p1.ads and p2.ads) but did not add them back. Could you double-check?

danielmercier: You are right! Will add them back...

clementfumex:
We approve the Ocaml code.

PR #292, opened 2019-07-29
nameres fixes & dispatching calls in find_all_refs

Roldak:


Roldak:
> Please add regression tests for both commits. Thanks!

Done! (among other things :D)

ada/language/ast.py
```diff
@@ -8209,7 +8296,16 @@ def find_all_references(units=AnalysisUnit.array):
         units.
         """
         dn = Var(Entity.canonical_part)
-        return dn.filter_is_imported_by(units, True).mapcat(
+
+        # If `dn` defines a subprogram which overrides some subprogram P, we
```
raph-amiard: I took a long time figuring that out so tell me if I'm wrong:

It's OK to take `base` instead of `dn` because every derived instance necessarily "withes"  the unit in which the parent is defined, and transitively up to the base. This wouldn't be true if `transitive` wasn't set to true below.

If that's correct, that might warrant an additional comment :)

ada/language/ast.py
```diff
@@ -1307,6 +1307,63 @@ def subp_decl_match_signature(other=T.BasicDecl.entity):
...
+        return Entity.is_subprogram.then(
+            lambda _: Entity.subp_spec_or_null.then(
+                lambda spec: spec.primitive_subp_of.find(
+                    lambda t:
+                    t.is_tagged_type | t.private_completion._.is_tagged_type
+                ).then(
+                    lambda t:
+                    t.primitives_env.get(Entity.name_symbol).filtermap(
+                        lambda bd: bd.cast(BasicDecl),
+                        lambda bd: bd.is_a(BasicDecl)
```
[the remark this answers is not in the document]
raph-amiard: Actually, that code is plain incorrect: You'll return every primitive with the same name even if it has a different profile:

```ada
procedure B is
   package P is
      type T is tagged null record;

      procedure Foo (Self : T; A : Integer) is null;
      procedure Foo (Self : T; A : Float) is null;


      type T2 is new T with null record;
      overriding procedure Foo (Self : T2; A : Float) is null;
      --% node.p_base_subp_declarations -> [
      --     <NullSubpDecl ["Foo"] b.adb:6:7-6:51>,
      --     <NullSubpDecl ["Foo"] b.adb:5:7-5:53>,
      --     <NullSubpDecl ["Foo"] b.adb:10:7-10:63>
      --  ]
   end P;
begin
   null;
end B;
```

ada/language/ast.py
```diff
@@ -3281,7 +3281,20 @@ def previous_part(go_to_incomplete=(Bool, True)):
                     pp.cast(T.BaseTypeDecl)._.is_private),
                 And(go_to_incomplete,
                     pp.is_a(T.IncompleteTypeDecl)),
-            )).cast(T.BaseTypeDecl))
+            )))._or(
+                # the previous part of this type is not in its semantic
+                # parents: this must mean we are in a package body and
+                # therefore the previous part must be in the private part of
+                # the package spec (RM 3.10.1).
```
raph-amiard: Wowow, quoting the RM and all :grin:. Two things:

1. The whole code you added seems unnecessary. This should be already handled by the above code (and it seems to be, since if I remove this chunk your test still passes).
2.  FIY, those types are commonly refered as "taft-amendment types". This would be good to add that to the comment, that will ring a bell more easily than a reference to the RM (which is a good thing to have anyway !)

ada/language/ast.py
```diff
@@ -3293,10 +3306,24 @@ def next_part():
...
             .get(itd.name.name_symbol, LK.flat, categories=noprims)
             .find(lambda t: t.is_a(BaseTypeDecl) & (t != Entity))
-            .cast(BaseTypeDecl),
+
+            # Or in the particular case where the incomplete decl is in the
+            # private part of the package spec, the next part can be found in
+            # the package's body (RM 3.10.1).
+            ._or(Entity.is_in_private_part.then(
+                lambda _:
+                Entity.parent.parent.parent
```
raph-amiard: `.parent.parent.parent` seriously :disappointed: . Try `.declarative_scope` instead.

ada/language/ast.py
```diff
@@ -3293,10 +3306,24 @@ def next_part():
...
+                .cast(T.BasePackageDecl).body_part.then(
```
raph-amiard: `.cast_or_raise`

ada/language/ast.py
```diff
@@ -3682,19 +3682,25 @@ def primitive_type_accessor():
         return T.PrimTypeAccessor.new(prim_type=Entity)
 
     @langkit_property(memoized=True)
+    def compute_primitives_env(include_self=(Bool, True)):
```
raph-amiard: FIY: stopping there, you already have lots of work to do with the previous comments. However: this seems to be severely lacking comments !

ada/language/ast.py
```diff
@@ -1307,6 +1307,63 @@ def subp_decl_match_signature(other=T.BasicDecl.entity):
...
+                    lambda t:
+                    t.primitives_env.get(Entity.name_symbol).filtermap(
+                        lambda bd: bd.cast(BasicDecl),
+                        lambda bd: bd.is_a(BasicDecl)
+                    )
+                )
+            )
+        )
+
+    @langkit_property(return_type=T.BasicDecl.entity, public=True,
```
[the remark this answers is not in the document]
Roldak: Done (`properties/root_subp_decls`)

ada/language/ast.py
```diff
@@ -1307,6 +1307,63 @@ def subp_decl_match_signature(other=T.BasicDecl.entity):
...
+        """
+        If Self declares a primitive subprogram of some tagged type T, return
+        the set of all subprogram declarations that it overrides (including
+        itself).
+        """
+        return Entity.is_subprogram.then(
+            lambda _: Entity.subp_spec_or_null.then(
+                lambda spec: spec.primitive_subp_of.find(
+                    lambda t:
+                    t.is_tagged_type | t.private_completion._.is_tagged_type
```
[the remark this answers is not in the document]
Roldak: Ok refactoring done

ada/language/ast.py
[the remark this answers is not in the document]
Roldak: Thanks for catching this! :D This should now be fixed, please have a look at the new implementation.

ada/testsuite/tests/properties/is_called_by/derived_2.ads
```diff
@@ -0,0 +1,5 @@
+with Derived;
```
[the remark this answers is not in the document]
Roldak: Done

ada/language/ast.py
Roldak: Yes that's it.
Done.

ada/language/ast.py
Roldak: Good catch, I removed this part.

ada/language/ast.py
Roldak: Okay I changed this part as well as the existing piece of code below.

ada/language/ast.py
Roldak: Done.

ada/language/ast.py
```diff
@@ -1317,8 +1317,7 @@ def base_subp_declarations():
         return Entity.is_subprogram.then(
             lambda _: Entity.subp_spec_or_null.then(
                 lambda spec: spec.primitive_subp_of.find(
-                    lambda t:
```
raph-amiard: You have another opportunity for simplification here:

https://github.com/AdaCore/libadalang/blob/master/ada/language/ast.py#L942

ada/language/ast.py
Roldak: Okay I added a comment but note that I didn't change any major logic here, this new property simply allows me to factor our a common piece of code (the part that does `group_env` with the explicit `with_md` argument) in both primitives_env and parent_primitives_env. (It was actually missing from `primitives_env` and so this refactoring allows fixing that conveniently)

ada/language/ast.py
Roldak: Actually I already implemented it (see https://github.com/AdaCore/libadalang/pull/292/commits/6620cfaa398c208171c49f76d99ee99812725e11) and I saw this opportunity but I then noticed that the semantics of the code you linked is not exactly the same as what I factored out. However if you think this change in semantics is okay, I can still proceed with the change.

raph-amiard: Ok fair enough

PR #293, opened 2019-07-30
Env hooks: do not crash on buggy name-less unit bodie

pmderodat:


pmderodat:
I agree this would be a very nice improvement! I also think we should not block this PR, as the refactoring you suggest may take some time… what do you think?

PR #300, opened 2019-08-12
Inline playground + FQN fix

raph-amiard:


ada/testsuite/tests/properties/fully_qualified_name_2/test.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import absolute_import, division, print_function
```
Roldak: Isn't this `test.py` useless if we use `driver: inline-playground`?

raph-amiard: It is! Good catch :)

PR #302, opened 2019-08-13
Preparatory work to unify all bare node types

pmderodat:


raph-amiard:
Is it expected that you have two commits with the same name ?

pmderodat:
Yes: one update is for the property name unification while the other is for the `use` clause refactoring in Langkit. So both have different tickets :-) Thanks for the reviews!

PR #303, opened 2019-08-13
S412-019: Implement get_primitives

Roldak:


ada/language/ast.py
```diff
@@ -492,6 +492,30 @@ def unique_units_impl(list_of_units=AnalysisUnit.array, i=Int,
             result
         )
 
+    @langkit_property(return_type=T.AdaNode.entity.array)
+    def unique_nodes(list_of_nodes=T.AdaNode.entity.array):
+        """
```
raph-amiard: This is gross.

raph-amiard: Ok let's merge it but you need to implement a `unique` method on arrays.

Roldak: Ok

PR #305, opened 2019-08-16
Rework get_aspect

raph-amiard:


Roldak:
Maybe we don't care about this but:
```ada
procedure Test is
   package Pkg is
      procedure Foo (X : Integer) is null;
      procedure Foo (X, Y : Integer) is null
         with Inline => False;

      pragma Inline (Foo);
   end Pkg;
begin
   null;
end Test;
```
Here `p_has_aspect('inline')` returns True for both Foo. (And needless to say, this code compiles fine)
I think a simple fix would be to look for `pragma`s in last instead of first.

Roldak:
I tested with `import` instead of `inline` and removed the `is null` of both procedures to make them basic decls:
 - with `Import => True` it says "cannot have two import pragmas", suggesting that this is not valid.
 - with `Import => False` it says "missing body", suggesting that it is valid and that the aspect has taken precedence because it now expects a body for Foo.
 - with `Import => False` and adding a body to it, we get a nice `GNAT BUG DETECTED` box.

So let's say we don't care about this :smile:

raph-amiard:
Thank you @Roldak for fuzzing the compiler!

PR #312, opened 2019-08-22
C API: do not emit warning for missing directories in loaded project

pmderodat:


pmderodat:
Yes, and that’s the point :-) Thanks!

PR #313, opened 2019-08-28
API changes process + actual API changes

raph-amiard:


changes/process_changes.py
```diff
@@ -0,0 +1,124 @@
+# coding: utf-8
```
pmderodat: Beyond testcases, none of our source files contain any non-ASCII character, so I would be in favor of removing this.

changes/process_changes.py
```diff
@@ -0,0 +1,124 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+import argparse as A
+from funcy import memoize
+from glob import glob
+import os
+from os import path as P
```
pmderodat: Nitpick: what about `import os.path as P`? Easier to read IMO

changes/process_changes.py
```diff
@@ -0,0 +1,124 @@
+# coding: utf-8
+
+from __future__ import absolute_import, division, print_function
+
+import argparse as A
+from funcy import memoize
```
pmderodat: We generally separate imports of third-party packages from standard library’s. Can you move `funcy` to a new block below?

changes/process_changes.py
```diff
@@ -0,0 +1,124 @@
...
+from os import path as P
+import re
+import yaml
+
+
+TN_RE = re.compile(r"[A-Z]\d\d\d-\d\d\d")
+
+
+def all_entries():
+    return [f for f in glob('*.yaml') if not f == 'entry_schema.yaml']
```
pmderodat: The condition looks wrong: should be `if f != 'entry_schema.yaml`

changes/process_changes.py
```diff
@@ -0,0 +1,124 @@
...
+import argparse as A
+from funcy import memoize
+from glob import glob
+import os
+from os import path as P
+import re
+import yaml
+
+
+TN_RE = re.compile(r"[A-Z]\d\d\d-\d\d\d")
```
pmderodat: This is unused: can you remove it? (it’s also wrong, as ticket numbers contain letters for Oct-Dec months)

changes/process_changes.py
```diff
@@ -0,0 +1,124 @@
...
+    for fname in all_entries():
+        tn = fname.split(".")[0]
+        with open(fname) as f:
+            entry = yaml.safe_load(f)
+            entry['tn'] = tn
+            validate_entry(tn, entry)
+            if entry['type'] == change_type:
+                entries.append(entry)
+
+    entries = list(sorted(entries, key=lambda e: e['date'], reverse=True))
```
pmderodat: No need to call `list`, as `sorted` already builds one.

changes/process_changes.py
```diff
@@ -0,0 +1,124 @@
...
+
+    print(header("Libadalang API {}".format(header_chunk), "#"))
+
+    for entry in entries:
+        print(header(entry["title"], "="))
+        print(entry["description"])
+        print(field("tn", entry["tn"]))
+        print(field("date", entry["date"]))
```
pmderodat: As discussed live, the date has no specific meaning (it’s the the date of merge in `stable` branches) and used only for sorting entries. So I think it should be left out of the output to avoid confusion with meaningful dates (date of merge in `master`/`stable`).

changes/process_changes.py
```diff
@@ -0,0 +1,124 @@
...
+        print(field("apis", format_apis(entry.get("apis"))))
+        print()
+
+
+@memoize
+def json_schema():
+    with open("entry_schema.yaml") as f:
```
pmderodat: Looks like you forgot to add this schema file: can you add it?

changes/process_changes.py
```diff
@@ -0,0 +1,124 @@
...
+
+types_to_header = {
+    'api-change': 'breaking changes',
+    'new-feature': 'new features'
+}
+
+
+def create_rst(change_type="api-change"):
+    """
+    Create an RST file from all the change entries for a given ``change_type``.
```
pmderodat: Currently, this function does not create an RST file (I mean does not write one to the disk) but rather prints RST content on the standard output. I we keep it that way, I suggest to update this docstring.

Correct me if I’m wrong, but for now the Sphinx doc will contain only breaking changes (not new features). Looks like this wasn’t the intent. What about we change the `create-rst` command to run `create_rst` once per change type? This would generate all sections (for all change types) in a single RST document, which sounds good to me.

changes/process_changes.py
```diff
@@ -0,0 +1,124 @@
...
+        with open(fname) as f:
+            entry = yaml.safe_load(f)
+            entry['tn'] = tn
+            validate_entry(tn, entry)
+
+
+if __name__ == '__main__':
+    os.chdir(P.dirname(P.abspath(__file__)))
+    parser = A.ArgumentParser()
+    subparsers = parser.add_subparsers()
```
pmderodat: Please add descriptions and help texts for the argument parser.

changes/process_changes.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    try:
+        from jsonschema import validate, ValidationError
+        try:
+            validate(entry, json_schema())
+        except ValidationError as e:
+            print("Error when validating entry for {}".format(tn))
+            print(e)
+    except ImportError:
+        pass
```
pmderodat: Given that we don’t do any validation if we can’t import `jsonschema`, I would suggest that we print a warning in that case.

changes/process_changes.py
```diff
@@ -0,0 +1,124 @@
...
+def validate():
+    """
+    Validate all yaml entries.
+    """
+    for fname in all_entries():
+        tn = fname.split(".")[0]
+        with open(fname) as f:
+            entry = yaml.safe_load(f)
+            entry['tn'] = tn
+            validate_entry(tn, entry)
```
pmderodat: This is mostly duplicate from the corresponding code in `create_rst`. What about moving the iteration/yaml loading/TN update parts in `all_entries`?

changes/S821-027.yaml
```diff
@@ -0,0 +1,8 @@
+type: api-change
+title: Rename ``BasicDecl.p_is_unit_root`` into ``p_is_compilation_unit_root``
+
+description: |
+    To be closer to common Ada terminology, and disambiguate between
```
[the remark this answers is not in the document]
raph-amiard: > Rename the p_is_unit_root property on BasicDecl into p_is_compilation_unit_root. This makes it closer to […]

I don't see it as more helpful. What do you think is more helpful ? that you phrase it as "`p_is_unit_root` property on `BasicDecl`" rather than "`BasicDecl.p_is_unit_root`" ? I think the dot notation is the only element in all of that that will be understandable by everyone, and it is the only thing you change.

"property" is langkit jargon too, and the naming convention is the python one, but we have to choose one for those changelogs I guess. I personally don't find your version more helpful. I also don't see the interest of repeating the description if the title already contains it. That won't make it more readable or more helpful, and repeating the same thing twice raises the risk of making a mistake.

So I would rather change the proposal to be more precise and say that the description doesn't have to be repeated.

changes/S821-027.yaml
```diff
@@ -0,0 +1,8 @@
+type: api-change
+title: Rename ``BasicDecl.p_is_unit_root`` into ``p_is_compilation_unit_root``
```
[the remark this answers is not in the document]
raph-amiard: That's a fun restriction :neutral_face:. That might force us to have nondescriptive titles sometimes I imagine. In light of that, my previous comment is kind of irrelevant: I don't see many cases where 52 characters will be enough to describe a change completely, not even this very simple case here.

ada/language/ast.py
```diff
@@ -10376,6 +10376,21 @@ def unit_kind():
...
+                unit_decl.has_aspect('Preelaborate'),
+                unit_decl.has_aspect('Shared_Passive'),
+                unit_decl.has_aspect('Remote_Types'),
+                unit_decl.has_aspect('Remote_Call_Interface'),
+            ))
+        )
+
     env_spec = EnvSpec(
         set_initial_env(Let(
             lambda n=Self.body.cast(T.LibraryItem).then(
```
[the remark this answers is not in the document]
raph-amiard: It does indeed! I only thought to document breaking changes so far, didn't update my brain into "document features too" mode.

changes/S828-005.yaml
```diff
@@ -0,0 +1,14 @@
+type: api-change
+title: Replace ``Name.p_referenced_id`` by ``Name.p_referenced_defining_name``
```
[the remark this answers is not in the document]
raph-amiard: How about, we strip them for impactdb? I prefer not to encumber our changes database with arbitrary restrictions, unless we have a good reason too.

changes/S821-025.yaml
```diff
@@ -0,0 +1,11 @@
+type: api-change
+title: Get rid of ``Name.p_xref``
+
+description: |
+    Its behavior is considered non intuitive, and is not properly
+    documented, nor tested.
+
+    The workaround is to replace calls to ``p_xref`` by the equivalent set of
+    navigation property calls.
```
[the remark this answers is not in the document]
raph-amiard: Ok

changes/process_changes.py
raph-amiard: OK

changes/process_changes.py
raph-amiard: Sure!

changes/process_changes.py
raph-amiard: OK

changes/process_changes.py
raph-amiard: Good catch, thanks, I wonder why it wasn't caught by the style checker ...

changes/process_changes.py
raph-amiard: Indeed, thanks!

changes/process_changes.py
raph-amiard: Hmm I wonder if I agree. In the current situation the date will be approximately the date of the merge in master, or at least will give a rough idea. This can still be useful to people, as it gives a lower bound about in which version of LAL you can expect the feature to be here. Otherwise, this info will be pretty hard to find.

changes/process_changes.py
raph-amiard: Good catch, thanks!

changes/process_changes.py
raph-amiard: > Looks like this wasn’t the intent

It was. The origin of all this is that I wanted an easy way for our users (mostly internal), and even for us, to glance at the breaking changes since last time they updated.

> This would generate all sections (for all change types) in a single RST document, which sounds good to me.

I wanted to have a different document/section per change type, to make it easier to browse. What do yo uthink ?

changes/S821-027.yaml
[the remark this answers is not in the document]
pmderodat: Ok

changes/S821-027.yaml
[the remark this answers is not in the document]
pmderodat: Yes.

changes/S828-005.yaml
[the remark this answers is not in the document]
pmderodat: Ok to strip them.

changes/process_changes.py
pmderodat: Because another module could do `process_changes.TN_RE`, so style checkers can’t detect this.

changes/process_changes.py
pmderodat: I’m only aware of AdaCore engineers who want to have this info, and they care only about the `stable` branch, for which we agreed to have a specific workflow (record the date of the branch bump during the bump IIRC), so I’m not convinced this will be useful in practice. My preference goes towards starting to be conservative and not show some approximate info.

changes/process_changes.py
pmderodat: Yes sorry this is what I meant to suggest: generate a single RST document that contains one section per change type, and then all change types listed there.

changes/process_changes.py
raph-amiard: I think you are wrong: AdaCore engineers care about the master branch, specifically about the diff of changes between current master and current stable, since usually they care about this list when we are going to bump stable. Could you clarify your reasoning or tell me if you missed something here ?

changes/process_changes.py
raph-amiard: Actually I think I'm even going to put it as a top level import. What do you think ?

pmderodat: Ok, but still protect catch ImportErrors and print warnings in this case, as jsonschema is not available in our production system.

changes/process_changes.py
pmderodat: I forgot about this use case, thanks.

Now, as I’ve written in private, we need to have data about which change appears in which release. We could use the same mechanism to track which change is in `stable` and which isn’t in any release (not even wavefronts). That will provide the info people want without showing approximate dates. Thoughts?

changes/process_changes.py
raph-amiard: Right

changes/process_changes.py
raph-amiard: I'll remove the date for now. Closing this and let's discuss this via a more synchronous medium :)

changes/process_changes.py
raph-amiard: SGTM. Actually I'll even move validation to the iterator.

changes/process_changes.py
raph-amiard: OK, done

user_manual/conf.py
```diff
@@ -352,6 +358,19 @@ def setup(app):
...
+    import subprocess
+    from os import path as P
+    import sys
+
+    rst_content = subprocess.check_output([
+        sys.executable, P.join("..", "changes", "process_changes.py"),
+        "create-rst"
+    ])
+
+    with open("{}/api_changes.rst".format(P.dirname(P.abspath(__file__))),
```
pmderodat: Using string formatting to create path just here while everything else uses `os.path` is odd: why not go all the way?

```
with open(P.join(P.dirname(P.abspath(__file__)), 'api_changes.rst'), 'w') as f:
```

ada/language/ast.py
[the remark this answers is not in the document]
pmderodat: The feature entry for this is still missing: can you add one?

PR #319, opened 2019-09-05
Minor fixes to process_changes.py

pmderodat:


changes/process_changes.py
```diff
@@ -5,14 +5,21 @@
...
 from glob import glob
 import os
 import os.path as P
+import sys
 import yaml
 
 from funcy import memoize
 
 
+def print_err(*args, **kwargs):
```
raph-amiard: No doc, also a bit convoluted way to write it, took me a while to understand that it's the same as

```py
print(*args, file=sys.stderr, **kwargs)
```

changes/process_changes.py
```diff
@@ -83,42 +90,42 @@ def header(title, header_char):
...
 
-    print(header("Libadalang API {}".format(header_chunk), "#"))
+    print(header('Libadalang API {}'.format(header_chunk), '#'))
 
     for entry in entries:
-        print(header(entry["title"], "="))
-        print(entry["description"])
-        print(field("tn", entry["tn"]))
-        print(field("apis", format_apis(entry.get("apis"))))
+        print(header(entry['title'], '='))
```
raph-amiard: I'm a bit troubled by your systematic replacement of double quotes by single quotes. Is there a reason/style guide you're following here ? There's nothing in PEP8 about it, nor in our own Python style guide.

In general if you feel strongly enough about a style convention to silently go over other people's code after the fact and modify it, it would be good to communicate about it.

changes/process_changes.py
pmderodat: Right, I forgot to add it: will do, thanks.

changes/process_changes.py
pmderodat: As discussed live: this is just for consistency. In the future we should decide on one style and automatically check it (at least on pure Python source files).

PR #323, opened 2019-09-10
Fix is_dispatching_call and is_called_by + small fixes

Roldak:


ada/language/ast.py
```diff
@@ -7615,7 +7615,12 @@ def is_dispatching_call():
                         # No need to check that the type of the expression
                         # is exactly the classwide type of the expected
                         # type, but simply that it is classwide.
-                        c.expr.expression_type.is_classwide
+                        c.expr.expression_type.then(
```
raph-amiard: Deserves an explanation about why you're checking the access type (same for the anonymous access type bit of code below)

PR #328, opened 2019-09-12
(no-tn-check) Fix binary archive uploading on Travis

reznikmm:
Fix binary archive uploading on Travis for MacOS X. We need this archive for Ada Language Server.


reznikmm:
Indeed! Thank you!

PR #335, opened 2019-09-18
S917-019: Fix primitives on incomplete type decls

Roldak:


ada/language/ast.py
```diff
@@ -3400,6 +3400,15 @@ def previous_part(go_to_incomplete=(Bool, True)):
             ))).cast(T.BaseTypeDecl)
         )
 
+    @langkit_property(return_type=T.BaseTypeDecl.entity)
+    def find_next_part_in(decl_part=T.DeclarativePart.entity):
```
raph-amiard: No doc

ada/language/ast.py
```diff
@@ -3400,6 +3400,15 @@ def previous_part(go_to_incomplete=(Bool, True)):
             ))).cast(T.BaseTypeDecl)
```
raph-amiard: Did you profile the perf before/after that ? Seems like replacing an env search by a linear search could have a perf impact ...

raph-amiard: Also what does this fix? specifically

ada/language/ast.py
```diff
@@ -9395,15 +9395,14 @@ def primitive_subp_of():
             Entity.returns._.singleton
         ))
 
-        return Self.unique_nodes(types.map(
```
raph-amiard: I had to read the code to verify that you were not modifying anything. Putting that in a "Minor: reformatting" commit would be nice :slightly_smiling_face:

raph-amiard: My bad, you are indeed modifying something :man_facepalming:

ada/language/ast.py
Roldak: No I haven't, but well functional correctness is more important right? This allows the change in the next commit to work. (When a primitive is declared on an apparent incomplete type, we need to fetch its next_part and attach the primitive to it instead (as IncompleteTypeDecl is not a TypeDecl), which does not work if next_part is based on env lookups as this whole thing is happening during PLE and envs are not fully populated at this point).

raph-amiard: Yes, functional correctness is definitely more important :) But knowing the perf impact would still be interesting.

Could you run a quick profile on a big project and see what gives ?

PR #336, opened 2019-09-18
Add a workaround after a Langkit fix regarding memoization

pmderodat:


ada/language/ast.py
```diff
@@ -7348,8 +7348,13 @@ def name_designated_type_env():
     @langkit_property(
         return_type=AnalysisUnit, external=True, uses_entity_info=False,
         uses_envs=False,
-        call_non_memoizable_because='Getting an analysis unit cannot appear'
-                                    ' in a memoized context'
+
+        # TODO (S917-027): re-enable this protection or remove it once we
+        # moved forward on memoization soundness issues in Langkit.
+        call_non_memoizable_because=(
+            False and
```
Roldak: That's weird :grimacing: Maybe at least update the doc of `call_non_memoizable_because` to make it accept a `bool` (currently, its type annotation is `str | None`)

pmderodat: Ah good point! Hmmm to avoid multiplying changes, what about if I just write `None and 'blah blah'`?

PR #338, opened 2019-09-18
Refactor find_all_refs and implement find_all_derived_types, find_all_overrides

Roldak:


ada/language/ast.py
```diff
@@ -1037,6 +1042,45 @@ def env_get_real_from_node(from_node=T.AdaNode):
             ): If(c.is_null, from_node, c)
         ))
 
+    @langkit_property(public=True, return_type=T.AdaNode.entity.array,
+                      dynamic_vars=[origin, default_imprecise_fallback()])
+    def find_all_in(x=T.AdaNode.entity, mode=FindAllMode):
```
raph-amiard: Please name `x` `root`, will be a bit clearer. Also `find_all_in` could be named `find_all_refs_in`, feels like you a word.

ada/language/ast.py
```diff
@@ -1037,6 +1042,45 @@ def env_get_real_from_node(from_node=T.AdaNode):
...
+
+                False
+            ),
+            x.singleton,
+            No(AdaNode.entity.array)
+        ))
+
+    @langkit_property(return_type=T.AdaNode.entity.array,
+                      dynamic_vars=[origin, default_imprecise_fallback()])
+    def find_all_driver(units=AnalysisUnit.array, mode=FindAllMode):
```
raph-amiard: `find_all_refs_driver`

ada/language/ast.py
Roldak: Ok for renaming `X` to `Root`, however, I renamed `find_all_refs_in` to `find_all_in` on purpose, because this is not anymore a procedure for finding _references_. It's just a tree traversal that you can use to collect whatever kind of node you need (references, accept statements, etc.)

ada/language/ast.py
Roldak: Same as above

PR #344, opened 2019-09-21
Fix #254: clarify documentation wrt. supported language versions

### LINKED ISSUE #254: documentation on supported Ada/python versions
On the homepage is written
```
Libadalang is a library that can be used from Ada and Python. Most of its code is Ada code, generated from the language specification that you can find in ada/language.
```
[Python has multiple versions](https://wiki.python.org/moin/Python2orPython3) and from this description it is not clear which python versions are supported.
Similar Ada also has multiple versions.
As a user, I would like to know 
* whether there are requirements on the version of Ada and python.
     e.g. is a minimal version required or can I even use python1.x?
* are versions no longer supported (like python 2.6.x or 1.x)?
* with which versions libadalang is tested.

raph-amiard:


README.md
```diff
@@ -31,9 +31,10 @@ issue](https://github.com/AdaCore/libadalang/issues/). We also gladly accept
 High level architecture
 -----------------------
 
-Libadalang is a library that can be used from Ada and Python. Most of its code
-is Ada code, generated from the language specification that you can find in
-[ada/language](ada/language).
+Libadalang is a library that can be used from Ada (2012) and Python 2, amongst
+other languages (we also have a C api, and an experimental OCaml API). Most of
```
pmderodat: Please put `api` in upper case.

PR #348, opened 2019-09-25
S924-047: Handle files with multiple units.

Roldak:


ada/extensions/nodes/ada_node/bodies
```diff
@@ -87,6 +89,11 @@ function Ada_Node_P_Filter_Is_Imported_By
    --  Ada.Text_IO instead, which allows the correct behavior of this whole
    --  routine.
 
+   function All_Compilation_Units_From
+     (Root : Bare_Ada_Node) return CU_Array;
+   --  Given a node that is the root of an analysis units, return all the
```
raph-amiard: typo: 'root of an analysis unit'

ada/extensions/nodes/ada_node/bodies
```diff
@@ -55,6 +55,8 @@ function Ada_Node_P_Filter_Is_Imported_By
       Hash            => Hash,
       Equivalent_Keys => "=");
 
+   type CU_Array is array (Positive range <>) of Bare_Compilation_Unit;
```
raph-amiard: Maybe not important for now, but wondering if `All_Compilation_Units` shouldn't be defined as a regular property

PR #349, opened 2019-09-26
Implement `Globals` aspect resolution

raph-amiard:


ada/language/ast.py
```diff
@@ -252,12 +252,21 @@ def custom_id_text():
...
             lambda a: a.id.as_bare_entity.name_symbol.any_of('Pre', 'Post')
         )
     ).is_null))
 
+    @langkit_property()
+    def in_aspect(name=T.Symbol):
+        """
+        Return whether Self is contained by an aspect whose name is ``name``.
+        """
+        return Self.parents.find(
```
Roldak: Replace by `Self.parents.any(...)` maybe?

ada/language/ast.py
```diff
@@ -252,12 +252,21 @@ def custom_id_text():
         """
         return String("")
 
-    in_aspect = Property(Not(Self.parents.find(
+    in_prepost = Property(Not(Self.parents.find(
```
Roldak: It's a bit of a shame to write a generic `in_aspect` but not being able to use to implement this one :S

ada/language/ast.py
raph-amiard: Good point!

ada/language/ast.py
raph-amiard: `in_prepost` does only one parent chain lookup this way. Also it's 2 lines of code, not exactly complex logic :slightly_smiling_face:

PR #353, opened 2019-10-08
Fix memory leaks in OCaml bindings

danielmercier:
TN: S916-052


pmderodat:
Thanks for the fixes! I only have a minor comment: could you edit the commit message to mention that is relates to OCaml bindings? So, something like:
> Fix memory leaks in OCaml bindings

PR #365, opened 2019-10-31
Nameres stuff + array slice stuff

Roldak:


ada/language/ast.py
```diff
@@ -8158,6 +8162,36 @@ def designated_type_impl():
 
     params = Property(Entity.suffix.cast(T.AssocList))
 
+    @langkit_property(return_type=Bool, dynamic_vars=[origin])
+    def check_array_slice(typ=T.BaseTypeDecl.entity):
```
raph-amiard: `is_array_slice` would sound better and more coherent if this is a predicate

ada/language/ast.py
```diff
@@ -8192,6 +8192,16 @@ def check_array_slice(typ=T.BaseTypeDecl.entity):
             )
         )
 
+    @langkit_property(public=True, return_type=Bool)
```
raph-amiard: Just rename the `check_array_slice` into `is_array_slice` and give it a default origin.

Roldak: It's not just about origin, `check_array_slice` takes a parameter. But I agree the name is naze.

PR #367, opened 2019-11-07
Ada app framework

raph-amiard:


ada/extensions/src/libadalang-helpers.ads
```diff
@@ -0,0 +1,82 @@
...
+         --  Argument parser for your application. Supports a set of default
+         --  options. You can add your own on this parser.
+
+         package Charset is new Parse_Option
+           (Parser, "-C", "--charset", "Charset to use for source decoding",
+            Unbounded_String,
+            Default_Val => To_Unbounded_String ("iso-8859-1"));
+
+         package Scenario_Vars is new Parse_Option_List
+           (Parser, Short => "-X", Long => "--scenario-variables",
```
pmderodat: Do we have to provide a long option? This one is non standard (I don’t think we have long options for scenario vars), and if we preserve it, it should probably be `--scenario-variable` (one option = one scenario variable).

ada/extensions/src/libadalang-helpers.ads
```diff
@@ -0,0 +1,82 @@
...
+         package Project_File is new Parse_Option
+           (Parser, "-P", "--project",
+            Arg_Type    => Unbounded_String,
+            Default_Val => Null_Unbounded_String,
+            Help        => "Project file to use");
+
+         package Files_From_Project is new Parse_Flag
+           (Parser,
+            Long => "--files-from-project",
+            Help => "Take files from specified project file");
```
pmderodat: To avoid ambiguities, I suggest:
> Process files from the given project file (root project only).

ada/extensions/src/libadalang-helpers.ads
```diff
@@ -0,0 +1,82 @@
...
+            Name        => "files",
+            Arg_Type    => Unbounded_String,
+            Help        => "Files to analyze",
+            Allow_Empty => True);
+
+      end Args;
+
+      procedure Run;
+   end App;
+
```
pmderodat: Looks like a frequent use case will be to instantiate this generic and add command-line arguments: in that case, I think it would be helpful to add a small example that demonstrates the typical use case.

ada/extensions/src/libadalang-helpers.adb
```diff
@@ -0,0 +1,122 @@
...
+                        if FI.Language = "ada" then
+                           Files.Append (+Name);
+                        end if;
+                     end;
+                  end loop;
+                  Unchecked_Free (List);
+
+               end if;
+            end;
+         end if;
```
pmderodat: I think we should terminate with an error if `--files-from-project` is passed while no project file is passed.

ada/extensions/src/libadalang-helpers.adb
```diff
@@ -0,0 +1,122 @@
...
+         if not Args.Parser.Parse then
+            return;
+         end if;
+
+         Ctx := Create_Context
+           (Charset       => +Args.Charset.Get,
+            Unit_Provider => UFP);
+
+         --  Get explicitly passed files if necessary
+         if not Args.Files_From_Project.Get then
```
pmderodat: I suggest to terminate with an error if both `--files-from-project` and additional files are passed on the command-line (otherwise one will get silently discarded, error prone for users).

ada/extensions/src/libadalang-helpers.adb
```diff
@@ -0,0 +1,122 @@
...
+            begin
+               Initialize (Env);
+
+               --  Set scenario variables
+               for Assoc of Args.Scenario_Vars.Get loop
+                  declare
+                     A        : constant String := +Assoc;
+                     Eq_Index : Natural := A'First;
+                  begin
+                     while Eq_Index <= A'Length
```
pmderodat: This should be `Eq_Index <= A'Last`

ada/extensions/src/libadalang-helpers.adb
```diff
@@ -0,0 +1,122 @@
...
+               end loop;
+
+               Load (Project.all, Create (+Filename), Env);
+               UFP := Create_Project_Unit_Provider_Reference (Project, Env);
+
+               if Args.Files_From_Project.Get then
+
+                  List := Project.Root_Project.Source_Files;
+
+                  Sort (List.all);
```
pmderodat: Please add a comment:
> Get a sorted list of source files to get deterministic execution
(or similar)

ada/extensions/src/libadalang-doc_utils.adb
```diff
@@ -209,9 +209,9 @@ package body Libadalang.Doc_Utils is
    function Get_Documentation (Decl : Basic_Decl) return Doc_Type is
       Doc : Doc_Type;
    begin
-      if Decl.Kind in Ada_Base_Package_Decl
-        and then Decl.Kind /= Ada_Generic_Package_Decl
-      then
+      if Decl.Kind = Ada_Generic_Package_Internal then
+         return Get_Documentation (Decl.Parent.As_Basic_Decl);
+      elsif Decl.Kind in Ada_Base_Package_Decl | Ada_Generic_Package_Decl then
```
pmderodat: Looks like this change deserves a ticket, a changes entry and a testcase: could you take care of this?

user_manual/ada_api_tutorial.rst
```diff
@@ -9,6 +9,11 @@ actually do some practice with the Ada API.
 Preliminary setup
 =================
 
+.. attention:: This whole section is going to show you how to create a
+    Libadalang app from scratch. This is good, and walks you over the concepts
+    about how to use Libadalang. However, if all you want is a command line
+    generic application, you should consider using the :ref:`ada-generic-app`.
+
```
pmderodat: Very nice

user_manual/ada_api_tutorial.rst
```diff
@@ -396,3 +401,92 @@ name resolution in Ada:
...
+In order to facilitate the creation of Ada command line applications,
+we have added an ``App`` generic package, that you can simply
+instantiate in order to create a command line application with a lot
+of common functionality already built-in, so that you don't have to reinvent it
+every time.
+
+The way it works is that it handles all the logic for you. You just pass it one
+or two callbacks:
+
+- One callback, `Process_Unit`, will be called for each new `AnalysisUnit` that
```
pmderodat: This tutorial is Ada-specific, so I suggest to spell `Analysis_Unit` (same below). Also, this is Sphinx, so code should be surrounded with double backquotes (here and several times below).

user_manual/ada_api_tutorial.rst
```diff
@@ -396,3 +401,92 @@ name resolution in Ada:
...
+    procedure Main is
+    begin
+       App.App.Run;
+    end Main;
+
+Then, running the app on a project is as simple as
+
+.. code:: bash
+
+    # files are automatically deduced from the project file
```
pmderodat: Please capitalize comments

ada/extensions/src/libadalang-helpers.ads
```diff
@@ -0,0 +1,82 @@
...
+   generic
+      Description : String;
+
+      with procedure Process_Context
+        (Ctx : Analysis_Context; Units : Unit_Vectors.Vector) is null;
+      --  This procedure will be called once after all units have been parsed
+
+      with procedure Process_Unit
+        (Unit : Analysis_Unit) is null;
+      --  This procedure will be called once right after an unit is parsed
```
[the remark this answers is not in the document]
raph-amiard: Yes that's correct, and np I can swap them

ada/extensions/src/libadalang-helpers.ads
raph-amiard: You're right. I'll check `GNATCOLL.Opt_Parse` to see if we have to provide the long or not.

raph-amiard: So, I checked in `Opt_Parse`, and the long flag is indeed not optional. I think the reasoning at the time is that all flags should have a long descriptive form.

We can change that if you think it's necessary (I told you Opt_Parse was not yet stable! :), personally since this is backwards compatible I don't really care. I'll remove the s for this PR and we can see later for the bigger project of removing long if you think it's worth it.

ada/extensions/src/libadalang-helpers.ads
raph-amiard: Where, in the .ads?

I hadn't had to do it so far for my uses of `App`, and that's already kind of part of the documentation of `GNATCOLL.Opt_Parse`. I would propose to do it in the tutorial, at a later stage

ada/extensions/src/libadalang-helpers.adb
raph-amiard: Right, good catch, thanks!

raph-amiard: Although I'm starting to wonder... If you just pass a project file and no explicit files, `files-from-project` mode will be used. So is it worth keeping the explicit command line option at all?

ada/extensions/src/libadalang-helpers.adb
raph-amiard: That would seem to go in the direction of my previous comment ...

user_manual/ada_api_tutorial.rst
raph-amiard: Thanks!

ada/extensions/src/libadalang-helpers.adb
raph-amiard: As agreed live, I'll remove the option

ada/extensions/src/libadalang-helpers.adb
raph-amiard: As agreed live, I'll remove the option

ada/extensions/src/libadalang-helpers.ads
pmderodat: Keeping the long form is fine for now, IMO. Thanks!

ada/extensions/src/libadalang-helpers.ads
pmderodat: Having it in the tutorial SGTM

ada/extensions/src/libadalang-helpers.adb
```diff
@@ -0,0 +1,126 @@
...
+         --  Setup traces from config file
+         GNATCOLL.Traces.Parse_Config_File;
+
+         if not Args.Parser.Parse then
+            return;
+         end if;
+
+         Ctx := Create_Context
+           (Charset       => +Args.Charset.Get,
+            Unit_Provider => UFP);
```
pmderodat: This creates the context before the unit provider is created: IIUC as a result the context will never use the UFP we created below. Can you confirm? If so, it looks like just moving the initalization of Ctx right before iterating of Files should do the job.

user_manual/changes/SB06-030.yaml
```diff
@@ -0,0 +1,11 @@
+type: new-feature
+title: Add a command line app framework for Ada API
+
+description: |
+    Add a new package `Libadalang.Helpers`, with a generic package `App`, that
```
pmderodat: This field contains ReST-formatted text, so the backquotes should be doubled.

user_manual/ada_api_tutorial.rst
```diff
@@ -396,3 +401,92 @@ name resolution in Ada:
...
+of common functionality already built-in, so that you don't have to reinvent it
+every time.
+
+The way it works is that it handles all the logic for you. You just pass it one
+or two callbacks:
+
+- One callback, ``Process_Unit``, will be called for each new ``Analysis_Unit``
+  that the application processes.
+- One callback, ``Process_Context``, will be called at the end when every
+  AnalysisUnit has been processed.
```
pmderodat:
```
s/AnalysisUnit/``Analysis_Unit``/
```

user_manual/changes/SB08-016.yaml
```diff
@@ -0,0 +1,9 @@
+type: bugfix
+title: Fix ``BasicDecl.doc`` not working on generic packages
+
+description: |
+    The ``BasicDecl.doc`` was not working correctly on generic packages,
```
pmderodat: I have the feeling that you forgot a word: “the ``BasicDecl.doc`` property”

ada/extensions/src/libadalang-helpers.adb
raph-amiard: Hmmm, looks fishy indeed, and since my test doesn't use name resolution *and* all sources are in CWD, it probably wasn't caught.

Wondering if I should add a test for that, but not sure what it should look like ...

user_manual/changes/SB06-030.yaml
raph-amiard: Thanks

ada/extensions/src/libadalang-helpers.adb
raph-amiard: More specifically the Ada test driver is bad at testing this kind of thing, and I don't really want to add a driver just for that case. Since I plan on using this generic in `nameres`, maybe we can consider the provider will be tested as part of that?

pmderodat: Good idea, that would work for me.

PR #372, opened 2019-11-21
SB21-019: Handle access to subprograms in p_real_type.

Roldak:


ada/language/ast.py
```diff
@@ -2454,7 +2454,9 @@ def real_type(typ=T.BaseTypeDecl.entity):
             # Handle the case where the primitive is defined on an anonymous
             # access type, by returning an anonymous access type over the
             # real_type of the accessed type.
-            typ.is_a(AnonymousTypeDecl),
+            typ.cast(AnonymousTypeDecl).then(
+                lambda td: Not(td.is_subprogram_access)
```
raph-amiard: I wonder about the usefulness of your nine lines `is_subprogram_access` helper method, if you're just using here instead of doing 

```
Not(typ.cast(AnonymousTypeDecl)._.type_def.cast(AccessToSubpDef).is_null)
```

I'd suggest checking all the occurences of `AccessToSubpDef` to see if this helper method would be useful elsewhere in `ast.py`. Also, maybe hosting it on `BaseTypeDecl` and making it public would justify having an helper. As-is, I don't really think so.

As a general rule, I do not introduce helper methods if they're used only once, or in other words, follow the "factorization" rule :slightly_smiling_face:

Roldak: I first wanted to implement it as you suggest, the problem is that the correct way to write this would be:
```
typ.cast(AnonymousTypeDecl).then(
    lambda td: td.type_def.is_a(AccessToSubpDef)
)
```
Because in your suggestion it will return true for `typ`s that are not `AnonymousTypeDecl`s. (Due to the `Not` being around the underscore expansion).

I generally agree with what you said but IMO in some cases it's worth adding a property just to make the code cleaner and more readable. I can still inline this if you prefer.

raph-amiard: Right, your way is better, using `is_a`, I don't know why I didn't think of it - except your forgot the `Not` :) . You can still use the underscore notation in your example I think: 

`typ.cast(AnonymousTypeDecl)._.type_def.is_a(AccessToSubpDef)`

Yeah, I think having a property just to abstract a kind check is unnecessary

Roldak: Yea sorry I edited, but I think you forgot it too :P

raph-amiard: Lies!

user_manual/changes/SB21-019.yaml
```diff
@@ -0,0 +1,12 @@
+type: bugfix
+title: Navigation of subprograms with anonymous access parameters
+short_title: Subprograms with anonymous access parameters
+
+description: |
+    Navigation, for example through p_canonical_part, was broken when
```
raph-amiard: Plz use sphinx, notably use double backtick blocks when mentioning a source entity

PR #373, opened 2019-11-25
Migrate all test programs to Libadalang.Helpers.App

pmderodat:


ada/extensions/src/libadalang-helpers.ads
```diff
@@ -48,15 +49,20 @@ package Libadalang.Helpers is
       Description : String;
       --  Description for the application. Will be used in the help string.
 
-      with procedure Process_Context_Before (Ctx : Analysis_Context) is null;
-      --  This procedure will be called right before going through units
+      with procedure Process_Context_Before
```
raph-amiard: Originally, I didn't put the `Context` in the `Process_Unit` callback because you can reach it from the `Unit`. But if we're starting to have more `Context`, I wonder if we shouldn't put it in a record and pass it to every callback, because now users won't have access to the project tree from the `Process_Unit` callback.

Note: Just a thought, not review blocking

ada/extensions/src/libadalang-helpers.ads
```diff
@@ -34,6 +34,11 @@ package Libadalang.Helpers is
    package Unit_Vectors is new
      Ada.Containers.Vectors (Positive, Analysis_Unit);
 
+   Abort_App_Exception : exception;
+   procedure Abort_App (Message : String := "") with No_Return;
```
raph-amiard: This design is a bit strange: I would have expected either:

1. just a procedure Abort_App that takes a message (and a private hidden exception as an implementation detail).
2. just an exception, which when raised will be caught by the app framework, and print the embedded message on stderr before exiting.

Having both is a bit confusing for no reason IMHO. There's not much use in using the exception directly if you have the procedure.

ada/extensions/src/libadalang-helpers.adb
```diff
@@ -36,6 +36,29 @@ package body Libadalang.Helpers is
                  renames To_Unbounded_String;
    function "+" (S : Unbounded_String) return String renames To_String;
 
+   procedure Print_Error (Message : String);
+   --  Shortcut for Put_Line (Standard_Error, Message)
```
raph-amiard: According to the below implem, the comment here is misleading: the implem is more than just a shortcut for `Put_Line (Standard_Error, Message)`. What about just saying `Helper to print an error on stderr` if you don't want to get into the details.

ada/extensions/src/libadalang-helpers.ads
```diff
@@ -54,21 +85,45 @@ package Libadalang.Helpers is
...
-        (Unit : Analysis_Unit) is null;
+        (Context : App_Job_Context; Unit : Analysis_Unit) is null;
       --  This procedure will be called once right after a unit is parsed
 
-      with procedure Process_Context_After
-        (Ctx     : Analysis_Context;
-         Project : GNATCOLL.Projects.Project_Tree_Access;
-         Units   : Unit_Vectors.Vector) is null;
-      --  This procedure will be called once after all units have been parsed
+      with procedure Job_Tear_Down (Context : App_Job_Context) is null;
```
raph-amiard: I'm not a fan of the `Tear_Down` name, it's not like this could only be used for tear down actions, it's also the perfect place to do context wide actions (if you have an app that works at the global crossref graph level for example).

ada/extensions/src/libadalang-helpers.ads
```diff
@@ -54,21 +85,45 @@ package Libadalang.Helpers is
...
+      --  the project is loaded (if present) and the list of files to process
+      --  is computed.
+
+      with procedure Job_Setup (Context : App_Job_Context) is null;
       --  This procedure will be called right before going through units.  If a
       --  project was loaded, Project refers to it, otherwise it is null.
 
       with procedure Process_Unit
-        (Unit : Analysis_Unit) is null;
+        (Context : App_Job_Context; Unit : Analysis_Unit) is null;
```
raph-amiard: OK, so this obsoletes my previous comment :)

ada/extensions/src/libadalang-helpers.ads
pmderodat: Fair point. I’ll go for the procedure-only way (easier to use, and no restriction on Message). Thanks!

ada/extensions/src/libadalang-helpers.adb
pmderodat: Right. It used to be a real shortcut, but I later noticed that GNATCOLL.Projects.Load already includes line breaks in its error message through its Error callback… and Ada.Text_IO does not play nicely with that… Anyway, I’ll update the comment, thanks!

ada/extensions/src/libadalang-helpers.ads
pmderodat: After a live discussion: let’s rename that to `Job_Post_Process`.

ada/extensions/src/libadalang-helpers.ads
pmderodat: :-)

PR #375, opened 2019-11-25
SB25-043 Implement name resolution for pragma Volatile

danielmercier:
Associated entity of a pragma Volatile is the first argument.

### DIFF
--- a/ada/testsuite/tests/properties/get_volatile_aspect/test.adb
+++ b/ada/testsuite/tests/properties/get_volatile_aspect/test.adb
@@ -1,0 +1,9 @@
+procedure Test is
+   type VolatileType is new Integer;
+   pragma Volatile (VolatileType);
+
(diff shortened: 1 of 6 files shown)


pmderodat:
Hello Daniel,

Thanks for the fix! Could you add a testcase (in `ada/testsuite/tests/name_resolution`)? Also, we now document all new features and bug fixes: could you please add an entry in <https://github.com/AdaCore/libadalang/tree/master/user_manual/changes>? Thanks in advance!

ada/testsuite/tests/properties/get_volatile_aspect/test.adb
```diff
@@ -0,0 +1,9 @@
+procedure Test is
+   type VolatileType is new Integer;
```
raph-amiard: I'm not going to block the PR for naming convention in a test but: EWWWW

PR #386, opened 2019-12-04
SC04-030: Improve xref precision on primitive calls.

Roldak:


ada/language/ast.py
```diff
@@ -2607,10 +2607,8 @@ def defining_env():
             # Same for "derived" and "interface" type definitions, but we also
             # need to include the defining environments of their base types.
             Self.is_a(T.DerivedTypeDef, T.InterfaceTypeDef),
-            Entity.base_types.map(
-                lambda bt: bt._.defining_env
-            ).concat(
-                Array([Entity.children_env, Entity.previous_part_env])
+            Array([Entity.children_env, Entity.previous_part_env]).concat(
```
raph-amiard: Adding a comment explaining why it is necessary to concat in that order would seem good!

PR #387, opened 2019-12-05
Add `BaseTypeDecl.p_discrete_range` property

raph-amiard:


raph-amiard:
Sorry, forgot to add it, should be good now

PR #406, opened 2019-12-18
Topic/formal procedures nameres

raph-amiard:


ada/language/ast.py
```diff
@@ -607,6 +607,11 @@ def resolve_names_internal(initial=Bool, additional_equation=Equation):
         Designates entities that are entry point for the xref solving
         infrastructure. If this returns true, then resolve_names can be called
         on it.
+
+        .. note::
+            For convenience, and unlike what is defined in the ARM wrt.
+            complete contexts for name resolution, xref_entry_points can be
```
pmderodat: Please add code quotes around `xref_entry_points`

raph-amiard: Done

raph-amiard:
@Roldak This is wrong: the `Test_Statement` will only resolve the node immediately above. If you run the test with `nameres --all` however, you'll see that everything is resolved correctly. The other names to resolves are part of other xref entry points (hence my clarification comment on `xref_entry_point` in `ast.py`à

PR #410, opened 2019-12-19
Port find-all-references testcases to nameres

pmderodat:


ada/testsuite/ada/nameres.adb
```diff
@@ -52,9 +52,26 @@ procedure Nameres is
       --  Whether to display short images for resolved names
    end record;
 
+   type Reference_Kind is (Any, Subp_Call, Subp_Overriding, Type_Derivation);
+   --  Kind of reference for Find_All_References (FAR) pragmas. Each kind leads
```
raph-amiard: Funny, I just read an article about Win16 programming, where FAR was an acronym for something else https://en.wikipedia.org/wiki/Far_pointer. I'm not a fan of this acronym, I'd rather you use `Refs_Request` which is clearer and mostly as short. Less terminology

ada/testsuite/ada/nameres.adb
```diff
@@ -895,6 +1083,178 @@ procedure Nameres is
...
+                         & Locator (Ref_Decl));
+            end;
+
+            Empty := False;
+         end Print_Ref;
+
+      begin
+         for DN of Target.P_Defining_Names loop
+            Put_Line
+              (What & DN.Debug_Text & " " & Locator (DN));
```
raph-amiard: Nit: useless line break

ada/testsuite/testsuite_support/name_resolution_driver.py
```diff
@@ -13,7 +13,8 @@ class NameResolutionDriver(BaseDriver):
 
     @property
     def run_python(self):
-        return self.py_runner.is_python_api_available
+        return (self.py_runner.is_python_api_available and
```
raph-amiard: I'd rather go all the way and get rid of it at this stage ... It introduces a lot of complexity, the duplication of a lot of logic, all of that for a very dubious purpose. What do you think ?

pmderodat: Agreed.

pmderodat:
As discussed live: this is clearly not ideal. In the future, we’ll manage to get rid of our Python/Valgrind interaction issues and port these testcases to the inline playground driver, making it possible to remove this complexity from `nameres.adb`. In the meantime, we’ll have Valgrind coverage for Find_All_References properties.

PR #427, opened 2020-01-22
S912-013: Work on P_Unit_Dependencies

Roldak:


ada/language/ast.py
```diff
@@ -11296,6 +11308,76 @@ def unit_kind():
...
+                kind=UnitSpecification
+            )._or(Self.designated_compilation_unit(
+                p.as_symbol_array,
+                kind=UnitBody
+            )).as_bare_entity
+        )
+
+    @langkit_property(return_type=T.CompilationUnit.entity.array,
+                      memoized=True)
+    def imported_units():
```
raph-amiard: If you make `withed_units` public, you can probably make this property public too

user_manual/changes/S912-013.yaml
```diff
@@ -0,0 +1,10 @@
+type: new-feature
+title: Add P_Unit_Dependencies property
```
raph-amiard: You documented the addition of this property but not `withed_units` (and `imported_units` if you make it public too)

PR #449, opened 2020-02-14
T203-023: Fix nameres of binop with operands of universal types.

Roldak:


ada/language/ast.py
```diff
@@ -7317,31 +7342,74 @@ def no_overload_equation():
...
+                # logic variables as parameters.
+
+                # note: in the relevant cases below, we call base_subtype as a
+                # conversion property because operators are defined on the root
+                # subtype, so the return value will always be of the root
+                # subtype.
+
+                # Case 1: Both operands have an universal type: set the result
+                # type to that universal type.
+                Predicate(BaseTypeDecl.is_universal_type,
```
raph-amiard: That's really dirty. I think this should be the last patch of this kind until we merge the new solver ...

PR #451, opened 2020-02-17
T211-029: Fix nameres of index constraints.

Roldak:


ada/language/ast.py
```diff
@@ -4544,9 +4544,17 @@ class IndexConstraint(Constraint):
...
-    xref_equation = Property(
-        Entity.constraints.logic_all(lambda c: c.xref_equation)
-    )
+    @langkit_property()
+    def xref_equation():
+        typ = Var(Entity.subtype)
+        return Entity.constraints.logic_all(
+            lambda i, c:
+            c.xref_equation
+            & c.cast(T.Expr).then(
```
raph-amiard: Need moar comments! Why are you casting to `Expr` and then doing a `then` ? What are you doing with the other cases ? What is that `type_bind_val` doing ?

What will your code do with the following snippet ? Will it work because the range is assigned the type `Character` explicitly ?

```ada
   subtype List_AE2 is List (Character range 'A' .. 'E');
```

PR #457, opened 2020-03-03
 T303-005: Rework output of fully_qualified_name.

Roldak:


ada/language/ast.py
```diff
@@ -1941,7 +1941,21 @@ def fully_qualified_name():
         """
         Return the fully qualified name corresponding to this declaration.
         """
-        return Entity.sym_join(Entity.fully_qualified_name_array, String("."))
+        return Entity.string_join(
+            Entity.fully_qualified_name_impl.map(lambda t: t.text),
+            String(".")
+        )
+
+    @langkit_property(public=True, return_type=T.String)
```
raph-amiard: This is an API change (actually two) so needs a change entry :slightly_smiling_face: 

Also, it's a nit, but the way you cut your work into commits is not great for review: the two previous commits don't make sense before you read the last one. In those cases, you need to either state why this change is needed in the commit message, or you can do just one commit (it's not like it would have been a big commit either way)

Roldak: Okay I'll keep that in mind!

Roldak: Also, I added the change entry

PR #458, opened 2020-03-03
T302-029: Address subtype indications in case constraints.

Roldak:


ada/language/ast.py
```diff
@@ -9235,7 +9235,10 @@ def xref_equation():
                 Self.type_bind_val(e.type_var, Self.expr.type_val)
                 & e.sub_equation,
 
-                # TODO: Bind other cases: SubtypeIndication and Range
+                # SubtypeIndication case (``when Color range Red .. Blue``)
+                lambda t=T.SubtypeIndication: t.xref_equation,
+
+                # TODO: Bind other cases: Range
```
raph-amiard: Is that TODO really up to date ? Looking at the grammar, it should be handled by the `Expr` branch above. Can you check and remove

Roldak: Done!

PR #466, opened 2020-03-24
T212-032: Add ``p_is_visible`` property.

Roldak:


user_manual/changes/T212-032.yaml
```diff
@@ -0,0 +1,14 @@
+type: new-feature
+title: Add ``p_is_public_decl`` property
+
+description: |
+    This property can be called on a BasicDecl to determine if this
+    declaration is public from the point of view of a given node.
```
pmderodat: Please add an empty line, for proper Sphinx rendering.

user_manual/changes/T212-032.yaml
```diff
@@ -0,0 +1,14 @@
+type: new-feature
+title: Add ``p_is_public_decl`` property
+
+description: |
+    This property can be called on a BasicDecl to determine if this
```
pmderodat: Please add double backquotes around BasicDecl

user_manual/changes/T212-032.yaml
```diff
@@ -0,0 +1,14 @@
...
+    declaration is public from the point of view of a given node.
+    .. ATTENTION::
+        We only take into account private-visibility here. This means for
+        example that this considers visible a local entity declared in a
+        subprogram w.r.t the subprogram's outer scope.
+        Hence, this property should only be used to check if an entity is
+        *private* from a given point of view, but not generally *visible*.
+
+date: 2020-03-42
```
pmderodat: Please fix the date

ada/language/ast.py
```diff
@@ -1558,6 +1558,39 @@ def should_ref_generic_formals():
 
     is_in_private_part = Property(Self.parent.parent.is_a(T.PrivatePart))
 
+    @langkit_property(return_type=Bool, public=True)
+    def is_public_decl(origin=T.AdaNode.entity):
+        """
+        Return whether this declaration is public from the point of view of
+        the given ``origin`` node.
+
+        .. ATTENTION::
```
raph-amiard: It seems to me that to make your `ATTENTION` admonition irrelevant you could have a property more like:

`def is_public(from=T.Unit)` which would avoid your problems of scope visibility because only working at the library level (there would still be a potential misunderstanding about with visibility).

I'm also a bit afraid that this property is too specific, and, frankly, too weird (having to put such an admonition is a bad sign ...). I wonder if a really complete implementation of a real `is_visible_from`, taking into account nesting, with visibililty, etc, wouldn't also allow Guillaume to implement what he needs, while being wildly more useful in the general case. What do you think ?

Roldak: I agree. I updated my PR, which now contains a first implementation of `is_visible`. Although it is not complete, it will report an error if the user tries to call it on an unsupported case instead of returning a wrong answer.

raph-amiard: Great, thanks Romain!

PR #468, opened 2020-03-25
RA22-015: Remove more macros in ast.py.

Roldak:


ada/testsuite/tests/dsl_unparse/test.py
```diff
@@ -0,0 +1,21 @@
...
+    os.path.abspath(os.curdir),
+    'lal.lkt'
+)
+unparse_script = 'to:{},lexer,grammar,nodes'.format(unparse_dest)
+
+os.chdir(os.environ['LIBADALANG_ROOTDIR'])
+
+status = subprocess.call([
+    'ada/manage.py', 'generate', '-P',
+    '--unparse-script', 'to:{},lexer,grammar,nodes'
```
pmderodat: I suggest you add the `-vnone` argument to avoid the noisy logging.

ada/testsuite/tests/dsl_unparse/test.py
```diff
@@ -0,0 +1,21 @@
...
+
+import os
+import subprocess
+
+
+tmpdir = os.path.abspath(os.curdir)
+unparse_dest = os.path.join(
+    os.path.abspath(os.curdir),
+    'lal.lkt'
+)
```
pmderodat: I believe you can replace all the above with just: `unparse_dest = os.path.abspath('lal.lkt')`

PR #477, opened 2020-04-02
T103-024: Resolve generic formal objects

Roldak:


ada/language/ast.py
```diff
@@ -519,13 +535,37 @@ def resolve_names():
...
+    @langkit_property(return_type=Bool)
+    def resolve_names_from_closest_entry_point():
+        """
+        Resolve names from the closest entry point up to this node.
+
+        .. note::
+            unlike ``resolve_names``, this will *not* trigger resolution of
+            every node with stop_resolution that lie in the sub-tree formed by
+            the closest entry point. It will only resolve those that are in the
+            path to resolving Self.
```
raph-amiard, line 545: I think I understand, nevertheless an example would be good here IMHO.

ada/language/ast.py
```diff
@@ -467,32 +467,48 @@ def sub_equation():
...
-        return i & j
+    @langkit_property(return_type=Bool, dynamic_vars=[env, origin])
+    def resolve_names_internal():
+        """
+        Resolves names for this node up to xref_entry_point boundaries.
+        """
+        return Entity.resolve_own_names & Entity.resolve_children_names
+
+    @langkit_property(return_type=Bool, dynamic_vars=[env, origin])
+    def resolve_names_internal_with_eq(additional_equation=Equation):
```
raph-amiard, line 500: I'm curious: is there a reason you decided to split the implem in two ?

ada/language/ast.py
```diff
@@ -9382,6 +9311,181 @@ class AggregateAssoc(BasicAssoc):
...
+        # in type_val.
+        td = Var(If(
+            Not(mra.is_null),
+            mra.typ,
+            Entity.base_aggregate.type_val.cast(BaseTypeDecl),
+        ))
+
+        atd = Var(td.array_def)
+
+        return Cond(
```
raph-amiard, line 9425: 1. You have a `Cond` where an `If` would do
2. Seems to me like at this stage you can inline `general_xref_equation` inside `xref_equation`

ada/language/ast.py
```diff
@@ -9382,6 +9311,181 @@ class AggregateAssoc(BasicAssoc):
...
+        # First, try to find all the discriminants matched by this assoc
+        discr_matches = Var(agg.matched_discriminants.filter(
+            lambda pm: pm.actual.assoc == Entity
+        ))
+
+        # If there are none, this assoc matches one or several components of
+        # the record, so gather them.
+        # WARNING: It is important to gather these components ONLY IF this
+        # association is not for specifying a discriminant. Indeed,
+        # discriminants can (and must) be resolved separately once the type of
```
raph-amiard, line 9448: I think the key here is that discriminants must be resolved **first** (not only separately), so that the call to `matched_components` doesn't infinitely recurse (since it needs the value of the discriminants) right ? This is not very clear here.

ada/extensions/src/libadalang-implementation-extensions.adb
```diff
@@ -513,10 +513,11 @@ package body Libadalang.Implementation.Extensions is
    -------------------------------
 
    function Expr_P_Eval_As_Int_In_Env
-     (Node : Bare_Expr; Env : Internal_Substitution_Array_Access)
```
raph-amiard, line 516: Aren't you missing a test here ?

Roldak: You're right, forgot to check it in!

ada/language/ast.py
Roldak: Yes: `resolve_names_internal` can take advantage of memoization because it calls resolve_own_name to resolve self's names. `resolve_names_internal_with_eq` cannot be memoized because the additional equation is not hashable, so here the work will always have to be done in full, even when called with `LogicTrue()`.

ada/language/ast.py
Roldak: Yes that's right, although the piece of logic here is not for resolving them first, it's really to allow them to be resolved separately from the rest which _enables_ the whole mechanism to resolve them first.

ada/language/ast.py
Roldak: Added more explanations!

ada/language/ast.py
Roldak: Done!

ada/extensions/src/libadalang-implementation-extensions.adb
Roldak: Done!

PR #487, opened 2020-04-24
T406-026: Add P_Shapes property.

Roldak:


user_manual/changes/T406-026.yaml
```diff
@@ -0,0 +1,42 @@
+type: new-feature
+title: Add ``P_Shapes`` property
+description: |
+    This property, when called on the ComponentList of a record definition,
+    will return the list of all possible shapes that a record value of this
+    type can take. Consider the following record definition:
+
+	.. code::
```
raph-amiard, line 8: Looks like you mistakenly used tabs here

Roldak: Indeed, thanks!

PR #496, opened 2020-04-29
process_changes.py: fix --quiet handling

pmderodat:


user_manual/changes/process_changes.py
```diff
@@ -139,56 +141,53 @@ def get_entries(*args):
...
     result = '\n'.join(buffr)
 
-    # We always generate the html, even if we don't need a preview, so that
-    # we check that the rst is correctly formatted.
+    # We always generate the HTML, even if we don't need a preview, so that we
+    # check that the reST is correctly formatted.
     html = publish_string(result, writer_name='html')
+    if not args.quiet:
+        print(result)
+        print(html)
```
raph-amiard, line 176: It's still wrong (and it's my fault): we're not supposed to print html, that's a remain from my own debugging ... Documentation generation will fail this way.

pmderodat: Good point, thanks! I’ve removed the extra `print` statement.

PR #497, opened 2020-05-12
T510-001: Rework filter_is_imported_by.

Roldak:
In particular, use the recently introduced p_imported_units property to
discover "with"ed units. This removes a duplicate logic we had for discovering
those, and removes a bug where body-only top-level subprograms defined as child
packages would be erroneously filtered by p_filter_is_imported_by.

### DIFF
--- a/user_manual/changes/T510-001.yaml
+++ b/user_manual/changes/T510-001.yaml
@@ -1,0 +1,7 @@
+type: bugfix
+title: Missed child units during ``find_all_references``
+description: |
+    Until now, child units defining a subprogram body at the top-level without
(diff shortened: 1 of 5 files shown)


user_manual/changes/T510-001.yaml
```diff
@@ -0,0 +1,7 @@
+type: bugfix
```
pmderodat, line 1: Please add formatting to find_all_references (double back quotes)

Roldak: Thanks, fixed!

PR #498, opened 2020-05-13
T406-026: Include discriminant conditions in shapes.

Roldak:


ada/language/ast.py
```diff
@@ -3030,12 +3030,21 @@ def xref_equation():
     xref_entry_point = Property(True)
 
 
+class DiscriminantCondition(Struct):
```
raph-amiard, line 3033: I don't much like the term `Condition` here, I think it's pretty vague. What about `DiscriminantValues` with docstring Represent a set of values (as a list of choices) for a discriminant.

ada/language/ast.py
```diff
@@ -3030,12 +3030,21 @@ def xref_equation():
...
+    condition = UserField(type=T.AlternativesList.entity)
+
+
 class Shape(Struct):
     """
     Represent one of the shapes that a variant record can have, as a list of
     the available components.
     """
     components = UserField(type=T.BaseFormalParamDecl.entity.array)
+    conditions = UserField(type=T.DiscriminantCondition.array)
```
raph-amiard, line 3047: `discriminant_values` ?

PR #502, opened 2020-05-20
T513-013: Implement python method Name.doc_name

raph-amiard:
Not implemented in Libadalang proper due to missing DSL features in
langkit.

Also remove python methode Name.full_name, not useful in the current
state of things.

### DIFF
--- a/ada/extensions/python
+++ b/ada/extensions/python
@@ -9,18 +9,35 @@ def token_match(self, other):
 
 
 @property
-def full_name(n):
+def doc_name(n):
     """
-    Return a nicely pretty printed name for any expr that is only formed of
-    DottedName and Identifier instances.
-    """
-    if isinstance(n, DottedName):
-        return "{}.{}".format(n.f_prefix.full_name, n.f_suffix.full_name)
-    elif isinstance(n, BaseId):
-        return n.text
-    else:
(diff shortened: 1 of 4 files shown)


ada/extensions/python
```diff
@@ -9,18 +9,35 @@ def token_match(self, other):
...
-    DottedName and Identifier instances.
+    Format this name to be a readable qualified name for the entity designated
+    by it. Meant to be used in documentation context.
+
+    If the entity is local, it will return the relative name. If it is
+    non-local, return the shortest qualified name not taking use clauses into
+    account.
+
+    .. NOTE:: This is a python specific method, because for the moment this is
+        not conveniently implementable directly as a libadalang property.
```
pmderodat, line 22: In the commit message, you say that this is not implementable in the DSL: have you tried implementing it as an external property? For now, this is not only not available in the Ada API, but we’ll also have to break the Python API (name convention) once we manage to turn it into a real property.

raph-amiard: > have you tried implementing it as an external property?

No and I don't really want to, it's more trouble than it's worth right now.

It's not documented in API docs, nor in our change files, so it's concretely not part of the public API, which makes me think that we can just break it later, but if you want I can specify this in the docstring.

PR #509, opened 2020-05-27
T527-038: Expose `p_discriminants_list` property.

Roldak:


ada/language/ast.py
```diff
@@ -4152,7 +4152,9 @@ def is_definite_subtype():
     )
 
     discriminants_list = AbstractProperty(
-        type=BaseFormalParamDecl.entity.array
+        type=BaseFormalParamDecl.entity.array,
+        doc="Return the list of all discriminants of this type.",
```
pmderodat, line 4156: Can you extend this documentation? Specify what is returned when there are no discriminants and when discriminants are unknown.

PR #511, opened 2020-05-29
Minor testsuite adjustments

pmderodat:


pmderodat:
The `--build` command-line option is generally needed only when a testsuite needs to run on non-native platforms. As to what `enable_cross_support` is, see https://e3-testsuite.readthedocs.io/en/latest/api_testsuite.html#testsuite-environment

PR #512, opened 2020-05-29
Use GitHub Actions Workflow to build on Mac OS X

reznikmm:
See T529-007 for details.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,99 @@
+on:
+  push:
+    branches:
+      - master
+      - stable
+name: Build
+jobs:
+  build:
+    runs-on: macos-latest
+    name: Build and deploy
+    strategy:
+      fail-fast: false
+      matrix: # Build debug and production 
+        debug: ['', 'debug']      # '' if production, 'debug' for debug
+    steps:
+      - name: Get libadalang
+        uses: actions/checkout@v2
+      - name: Get gprbuild for libgpr
+        uses: actions/checkout@v2
+        with:
+          repository: AdaCore/gprbuild
+          path: gprbuild
+      - name: Get gnatcoll core
+        uses: actions/checkout@v2
+        with:
+          repository: AdaCore/gnatcoll-core
+          path: gnatcoll-core
+      - name: Get gnatcoll iconv/gmp
+        uses: actions/checkout@v2
+        with:
+          repository: AdaCore/gnatcoll-bindings
+          path: gnatcoll-bindings
+      - name: Get langkit
+        uses: actions/checkout@v2
+        with:
+          repository: AdaCore/langkit
+          path: langkit
+          fetch-depth: 0  # all branches
+      - name: Try to checkout langkit's branch
+        run: (cd langkit; git checkout ${{ github.ref }} || true)
+      - name: Get GNAT Community 2020 toolchain
+        uses: ada-actions/toolchain@b55aaa7
+        with:
+          distrib: community
+      - name: Setup Python 3.7
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.7'
+      - name: Build Libadalang and its dependencies
+        run: |
+          set -x -e
+          export PROCESSORS=0
+          export prefix=/tmp/ADALIB_DIR
+          export GPR_PROJECT_PATH=$prefix/share/gpr
+          DEBUG=${{ matrix.debug }}  # Value is '' or 'debug'
+          pip install -r langkit/REQUIREMENTS.dev
+          pip install jsonschema
+          python langkit/setup.py install
+          make -C gprbuild prefix=$prefix BUILD=${DEBUG:-production} GPRBUILD_OPTIONS="-cargs:C -Dst_mtim=st_mtimespec -gargs" libgpr.build.{shared,static} libgpr.install.{shared,static}
+          BUILD=`echo $DEBUG| tr [a-z] [A-Z]`  # Convert to uppercase
+          make -C gnatcoll-core prefix=$prefix BUILD=${BUILD:-PROD} LIBRARY_TYPES="relocatable static" build install
+          python gnatcoll-bindings/iconv/setup.py build ${DEBUG:+--debug} -j0 --prefix=$prefix --library-types=relocatable,static
+          python gnatcoll-bindings/iconv/setup.py install
+          export CPATH=/usr/local/include
+          export LIBRARY_PATH=/usr/local/lib
+          export DYLD_LIBRARY_PATH=/usr/local/lib
+          python gnatcoll-bindings/gmp/setup.py build ${DEBUG:+--debug} -j0 --prefix=$prefix --library-types=relocatable,static
+          python gnatcoll-bindings/gmp/setup.py install
+          eval `langkit/scripts/manage.sh setenv`
+          BUILD=${DEBUG:+dev}  # Convert debug to dev
+          # Build libadalang static library
+          ada/manage.py generate -P
+          ada/manage.py --library-types static build --build-mode ${BUILD:-prod}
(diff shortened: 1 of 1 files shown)


reznikmm:
Note: appveyor error is accidental and isn't related to the patch.

.github/workflows/main.yml
```diff
@@ -0,0 +1,66 @@
...
+      - uses: actions/checkout@v2
+        with:
+          repository: AdaCore/gnatcoll-bindings
+          path: gnatcoll-bindings
+      - uses: actions/checkout@v2
+        with:
+          repository: AdaCore/langkit
+          path: langkit
+          fetch-depth: 0
+      - run: pushd langkit; git checkout ${{ github.ref }} || true; popd
```
raph-amiard, line 30: Could you please:

1. Use `cd` instead of the more obscure `pushd`/`popd` ?
2. Comment this line ? I'm not sure I understand it, even though I guess it's to checkout the corresponding branch ?

TIA

.github/workflows/main.yml
```diff
@@ -0,0 +1,66 @@
...
+      - uses: ada-actions/toolchain@b55aaa7
```
raph-amiard, line 31: Why `@<specific sha1>` instead of `@dev` like written in Fabien's docs ?

.github/workflows/main.yml
```diff
@@ -0,0 +1,66 @@
...
+          distrib: community
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.7'
+      - run: |
+          set -x -e
+          export ADALIB_DIR=/tmp/ADALIB_DIR
+          export GPR_PROJECT_PATH=$ADALIB_DIR/share/gpr
+          pip install -r langkit/REQUIREMENTS.dev
+          pip install jsonschema
```
raph-amiard, line 42: I think if I understand correctly you're using langkit's req file because LAL's is not adapted because it contains a dep on langkit master, right ? I think instead you should:

1. use libadalang's own `REQUIREMENTS.dev`
2. Edit it to remove the dep on langkit, which is annoying 98% of the time
2. Also add the jsonschema dep so you can remove it from here.

Are you OK to do those changes ?

.github/workflows/main.yml
```diff
@@ -0,0 +1,66 @@
...
+          python langkit/setup.py install
+          make -C gprbuild BUILD=production prefix=$ADALIB_DIR GPRBUILD_OPTIONS="-cargs:C -Dst_mtim=st_mtimespec -gargs" libgpr.build.shared libgpr.install.shared
+          make -C gnatcoll-core PROCESSORS=0 prefix=$ADALIB_DIR LIBRARY_TYPES=relocatable build install
```
raph-amiard, line 45: Can't you rather `export PROCESSORS=0` at the top ?

.github/workflows/main.yml
```diff
@@ -0,0 +1,66 @@
...
+          python gnatcoll-bindings/iconv/setup.py build --reconfigure -j0 --prefix=$ADALIB_DIR --library-types=relocatable
+          python gnatcoll-bindings/iconv/setup.py install
+          export CPATH=/usr/local/include
+          export LIBRARY_PATH=/usr/local/lib
+          python gnatcoll-bindings/gmp/setup.py build --reconfigure -j0 --prefix=$ADALIB_DIR --library-types=relocatable
+          python gnatcoll-bindings/gmp/setup.py install
+          eval `langkit/scripts/manage.sh setenv`
+          ada/manage.py generate -P
+          ada/manage.py build
+          ada/manage.py install $ADALIB_DIR
```
raph-amiard, line 55: I'm confused, are you building LAL relocatable, and only relocatable ? I thought the ALS needed the static one. Are you sure you're not rebuilding LAL in the ALS's GH actions here ?

.github/workflows/main.yml
reznikmm: `@dev` doesn't include gnat community 2020. There is another branch `ce2020`, but it is unpublished yet on GH marketplace, so SHA1 hash protects us from stealing Bintray API key by Fabien :smiley:

.github/workflows/main.yml
reznikmm: I will. This is a replacement of
```
(
    cd langkit
    python2 ../utils/travis-langkit-branch.py
)
```
It tries to download the same langkit branch as the pushed lal branch (works only for `stable` now).

.github/workflows/main.yml
reznikmm: I'm afraid, it's too complicated for me. I just took lines from `utils/travis-install.sh`
```
# Install requirements
pip install -r langkit/REQUIREMENTS.dev
pip install jsonschema
(cd langkit && python setup.py install)

```

.github/workflows/main.yml
pmderodat: Hello Maxim,

I’m not sure I understand this part: the `travis-langkit-branch.py` script is designed to select the appropriate langkit branch for all CI situations, the `stable` branch included. So in theory it should be appropriate for GitHub actions. If you have detected a configuration where it does not behave correctly, could you give us the details?

.github/workflows/main.yml
raph-amiard: Ah, ok :) Thanks for the explanation

.github/workflows/main.yml
raph-amiard: Ok makes sense, I'll take care of the change in that case :)

.github/workflows/main.yml
raph-amiard: Outdated

.github/workflows/main.yml
reznikmm: 1. I don't have `python2` I guess
2. `travis-langkit-branch.py` uses `TRAVIS_*` variables those not provided by GH

pmderodat: 1. The script works fine with Python3: actually the AppVeyor build already runs it with Python3.
2. This should not be a problem: the AppVeyor build just provides the `TRAVIS_*` variables, as renamings of the `APPVEYOR_*` ones (see `utils/appveyor.sh`).

Thanks,

reznikmm:
Status update since yesterday discussion:
 * I've added steps names as comments
 * replaced `pushd` with `(cd`
 * Added `export PROCESSORS=0`
 * Enabled `static` libraries (both kinds of libraries are build now to help us migrate the ALS to `static` in the feature)
 * DEBUG build doesn't work with `gnatcoll-core` because of error:
```
gnatcoll-vfs.ads:47:01: warning: pragma "ADA_05" ignored
```
 * ` ada/manage.py build --build-mode prod` doesn't work on Mac OS because of error
```
   [link library] libadalang.dylib
Undefined symbols for architecture x86_64:
  "___gmpz_add_ui", referenced from:
      _libadalang__expr_eval__expr_eval_in_env in libadalang-expr_eval.o
```

So I keep just only one working configuration `PROD` gnatcoll-code with `dev` LAL.

raph-amiard:
Hey @reznikmm !

Can you try disabling the relocatable build ? It might be what is causing problems in PROD mode.

For the warning, it might mean that gnatcoll's build is failing with GNAT CE 2020, so not a good thing either...

reznikmm:
No, I need relocatable libraries to build ALS right now. I would prefer to keep `dev` mode for Libadalang until we change ALS to static build or somebody finds the reason why LAL in `prod` mode doesn't build on MacOs .

.github/workflows/main.yml
reznikmm: This single line of code meets all my needs for now.

reznikmm:
I've fixed a few issues in gnatcoll, so now it builds fine in `debug` mode.

reznikmm:
Now gh action build 4 archives (prod, dev) x (static x relocatable).

PR #524, opened 2020-06-04
Avoid requiring root to build

GUIpsp:
This pr modifies the default behavior of the build scripts in order to avoid needing root.

### DIFF
--- a/install-lal-and-deps.sh
+++ b/install-lal-and-deps.sh
@@ -28,7 +28,7 @@ set -e
 export ADALIB_DIR=$HOME/adalib
 export TOOLS_DIR=$HOME/build_tools
 export INSTALL_DIR=$HOME/install
-export LIBRARY_PATH=/usr/local/lib
+export LIBRARY_PATH=$HOME/libadalang/lib
 
 # This script defers the real work to scripts that are used for travis
 # automated builds. You can see and customize those if you need.


GUIpsp:
I am unsure of how my small diff is causing the build to fail. I use pretty much this patch locally and it works

pmderodat:
Hello @GUIpsp 

First, the build failures in Travis CI and AppVeyor are indeed unrelated to your patch: they were due to issues on our side, now fixed.

I’m trying to understand your patch. It seems that setting the `LIBRARY_PATH` environment variable isn’t necessary currently, so I would be in favor of removing this line. However I don’t understand how setting it to `/usr/local/lib` makes the build require superuser permissions. What happens on your side?

GUIpsp:
You are indeed correct, and the patch as posted doesn't make a difference. The previous patch I had posted (https://github.com/AdaCore/libadalang/commit/a9107df6885e618cd88affd36f15aebeb32579e5) was the one that fixed me requiring root. However, that messes with venv so I'm not sure if there is a good way to support both user and venv installs

pmderodat:
Ah, I understand, thank you for the context. Using pip’s `--user` would likely solve the problem in your setup, but that indeed would break setups using virtualenv (for instance our Travis CI). The `install-lal-and-deps.sh` script is supposed to be a template for you to update locally at your convenience, put the calls to `pip` happen in our Travis scripts: do you want us to introduce an environment variable such as `PIP_USER_MODE` so that you can turn on and off pip’s `--user` option in our Travis scripts? I don’t think we can do much more at this stage.

GUIpsp:
I think I have I'll just update my setup to be venv based and save the trouble. Thank you for your time.

GUIpsp:
However, another minor issue is that running the script multiple times will complain due to already existing files. I think forcing can be the behavior for the script, and it shouldn't affect travis at all.

pmderodat:
You are welcome! I tried to reproduce the trouble with running the script multiple times, but that was successful. What is the error output you get?

GUIpsp:
The error is as follows (trimmed output):
```
Install project Libadalang - relocatable
file libadalang-doc_utils.ali exists, use -f to overwrite
Install failed: error while running gprinstall -p -P/home/guipsp/libadalang/build/lib/gnat/libadalang.gpr --prefix=/home/guipsp/adalib --build-var=LIBRARY_TYPE --build-var=LIBADALANG_LIBRARY_TYPE --sources-subdir=include/libadalang --build-name=relocatable -XBUILD_MODE=dev -XLIBRARY_TYPE=relocatable -XGPR_BUILD=relocatable -XXMLADA_BUILD=relocatable:
    Command '['gprinstall', '-p', '-P/home/guipsp/libadalang/build/lib/gnat/libadalang.gpr', '--prefix=/home/guipsp/adalib', '--build-var=LIBRARY_TYPE', '--build-var=LIBADALANG_LIBRARY_TYPE', '--sources-subdir=include/libadalang', '--build-name=relocatable', '-XBUILD_MODE=dev', '-XLIBRARY_TYPE=relocatable', '-XGPR_BUILD=relocatable', '-XXMLADA_BUILD=relocatable']' returned non-zero exit status 4.
```

PR #533, opened 2020-06-22
Minor reformatting & changes entry for T622-008

pmderodat:


ada/extensions/python_api/unit_providers/mypy_methods
```diff
@@ -0,0 +1,14 @@
+## vim: filetype=makopython
```
raph-amiard, line 1: This seems weird: I don't see any signs of the mypy_methods extension point anywhere: How and when did this get included ?

pmderodat: As discussed live: this is the  job of https://github.com/AdaCore/langkit/pull/380, and in the meantime this file will just be ignored. Thanks,

PR #535, opened 2020-06-24
T605-007: Fix if-expressions' type var bindings

raph-amiard:


pmderodat:
> This does not fix the original issue of the ticket

Interesting… could you extend the testcase so that actually checks the original issue?

raph-amiard:
Short update: This doesn't actually fix anything, just moves the problem somewhere else. We have some idea of how to fix it, but they're complicated and the problem is not a priority as of yet. Closing this PR.

PR #540, opened 2020-07-02
T701-019: Fix zip_with_params on incomplete type decls.

Roldak:


user_manual/changes/T701-019.yaml
```diff
@@ -0,0 +1,7 @@
+type: bugfix
+title: Fix ``zip_with_params`` on incomplete type decls
+description: |
+    This fixes an issue with calling ``P_Zip_With_Params`` on an discriminant
+    constraint that constrains the incomplete view of a discriminated type
+    would yield an empty result.
+title: 02-07-2020
```
pmderodat, line 7: Pasto (`title` should be `date`), and the date format is wrong (`2020-07-02`)

ada/language/ast.py
```diff
@@ -9913,7 +9913,7 @@ def zip_with_params():
             ._.formal_part.abstract_formal_params,
 
             lambda c=T.DiscriminantConstraint:
-            c.subtype.cast(T.TypeDecl)._.discriminants_list,
```
Roldak, line 9916: Not sure what this cast was for, I hope it won't break anything (it doesn't in our testsuite). Maybe you have an idea @raph-amiard ?

user_manual/changes/T701-019.yaml
Roldak: Thanks! I pushed too soon (testsuite wasn't done running :grimacing: ) Fixed!

ada/language/ast.py
raph-amiard: 1. I'm pretty sure the cast was wrong.
2. If it doesn't break anything in the testsuites, then I guess we'll see :)

PR #551, opened 2020-07-08
T708-003: Prevent name resolution in ``Convention`` aspect.

Roldak:


ada/language/ast.py
```diff
@@ -6379,6 +6379,11 @@ def xref_equation():
             # entry.
             Entity.id.name_is('Global'), Entity.expr.sub_equation,
 
+            # Do not resolve anything inside those aspect, as identifiers act
```
raph-amiard, line 6382: This won't work for `pragma Convention` right ?

Roldak: `pragma Convention` was already handled (see `Pragma.xref_equation`)

PR #560, opened 2020-07-21
T717-023: Also handle Modular_IO in special circuitry.

Roldak:


ada/testsuite/tests/name_resolution/text_io_nested_pkgs/test.adb
```diff
@@ -0,0 +1,9 @@
+with Ada.Text_IO;
+
+procedure Test is
+   type U is mod 10;
+   package MIO is new Ada.Text_IO.Modular_IO (U);
```
raph-amiard, line 5: Have you checked that there isn't any other package that should be handled by this special mechanism ?

Roldak: Yes, `Modular_IO` specifically was missing

PR #604, opened 2020-09-09
T909-028: Fix nameres equation for raise stmt's error message.

Roldak:


ada/language/ast.py
```diff
@@ -12641,7 +12641,18 @@ class RaiseExpr(Expr):
 
     @langkit_property()
     def xref_equation():
-        return Entity.exception_name.sub_equation
+        return And(
+            Entity.exception_name.sub_equation,
+            Entity.error_message.then(
+                lambda er: And(
+                    # The expected type of that error message is always String,
```
raph-amiard, line 12648: OMG you're starting to throw RM references all around; should I be worried ?

raph-amiard (approved):

ada/language/ast.py
Roldak: :sos: :sos: :sos:

PR #605, opened 2020-09-10
Various fixes

raph-amiard:


ada/testsuite/tests/contrib/check_subp_boxes/test.adb
```diff
@@ -0,0 +1,20 @@
+package body Test is
```
pmderodat, line 1: The commit message lacks a ticket number

ada/language/ast.py
```diff
@@ -11392,7 +11392,9 @@ def subp_decl_spec():
             Self.name_symbol, Self,
             dest_env=Entity.enum_type.primitives,
             metadata=T.Metadata.new(primitive=Entity.enum_type.node)
-        )
+        ),
+
+        add_env()
```
pmderodat, line 11397: Can you add a testcase for this? Also please add a comment, as it’s not clear why an env is needed here (EnumLiteralDecls can’t contain declarations, right?)

ada/language/ast.py
```diff
@@ -1968,14 +1968,14 @@ def constrain_prefix(prefix=T.Expr):
...
-        Entity.defining_name.relative_name, public=True, doc="""
+        Entity.defining_name._.relative_name, public=True, doc="""
         Return the relative name for Self. If Self's defining name is
         ``A.B.C``, return C as a node.
         """
     )
 
     relative_name_text = Property(
-        Entity.relative_name.symbol, doc="""
+        Entity.relative_name._.symbol, doc="""
```
pmderodat, line 1978: Please add a testcase.

Also, shouldn’t the commit message be “make p_relative_name return null when there is no defining_name”? (here it sounds like “make it aways return null”)

Thanks!

pmderodat:
Thanks a lot for taking care of SC11-035! Note that travis detected stylecheck issue, and the wrong branch of Langkit is used (needs your App changes).

PR #614, opened 2020-09-16
T915-004: Fix nameres of Mod attribute.

Roldak:


Roldak:
Actually I will include this commit in #615 since it's a dependency of it.

PR #620, opened 2020-09-28
Topic/t925 008

raph-amiard:


Roldak:
Cool! The only problem I have is that I can't see anywhere that the original issue from the ticket is fixed. Maybe in the testcase you could add a call to `p_is_dispatching_call` to check that it works as expected?

Roldak (dismissed):

Roldak (approved):

PR #623, opened 2020-09-29
Work around P_Base/Root_Subp_Declarations

Roldak:


ada/language/ast.py
```diff
@@ -2804,7 +2804,22 @@ def real_type(typ=T.BaseTypeDecl.entity):
...
-        canon_typ = Var(typ.then(lambda t: t.canonical_type.without_md))
+
+        canon_typ = Var(Cond(
+            typ.is_null,
+            No(BaseTypeDecl.entity),
+
+            # Don't call canonical_type on classwide types, as it returns a
+            # non-classwide type for them, which could make the comparison
+            # below incorrectly succeed.
+            # TODO: Fixing canonical_type directly would probably be better,
```
raph-amiard, line 2816: As discussed live: I'd like to have an idea of how hard it would be to do the real fix, rather than a workaround here

raph-amiard (changes requested):

ada/language/ast.py
Roldak: Actual fix done!

user_manual/changes/T925-015.yaml
```diff
@@ -0,0 +1,10 @@
+type: bugfix
+title: Fix output of ``P_Base/Root_Subp_Declarations``
+description: |
+    Until, those properties could return several subprogram declarations
```
raph-amiard, line 4: Until *now ?

raph-amiard (changes requested):

raph-amiard (approved):

PR #631, opened 2020-10-20
New laldoc

raph-amiard:


ada/testsuite/tests/contrib/laldoc/basic/test.out
```diff
@@ -0,0 +1,138 @@
+Generating rst
+==============
+
+Unknown annotation: unknown-annotation
+default entity handling for /home/amiard/libadalang/src/libadalang/tmp/contrib__laldoc__basic/base.ads:<SingleTaskDecl ["T"] base.ads:35:4-35:11>
```
Roldak, line 5: Not sure this should be part of the baseline :smile:

raph-amiard: You're right ... I should make the paths relative somehow

Roldak (approved):

PR #642, opened 2020-11-03
Changes & fixes needed for lkql/checkers

raph-amiard:


Roldak (dismissed):

extensions/src/libadalang-helpers.ads
```diff
@@ -41,6 +42,28 @@ package Libadalang.Helpers is
    package String_Vectors is new Ada.Containers.Vectors
      (Positive, Unbounded_String);
 
+   procedure Load_Project
+     (Project_File             : String;
+      Scenario_Vars            : Unbounded_String_Array := Empty_Array;
+      Target, RTS, Config_File : String := "";
+      Project                  : out Project_Tree_Access;
+      Env                      : out Project_Environment_Access);
+   --  Load Project_File using the given scenario variables
```
pmderodat, line 51: If this goes public, it is important to specify how the information is encoded in `Scenario_Vars`, and talk about error cases (for instance, what if there is no `=` in a string?)

pmderodat: Also, please Sphinx-ize comments so that they show up properly in the autodoc :)

pmderodat (approved):

PR #656, opened 2020-11-24
doinstall script: add unattended installation mode

pmderodat:


utils/doinstall.tmplt
```diff
@@ -157,6 +157,22 @@ check_env
 # is defined
 unset CDPATH
 
+if [ $# = 1 ]; then
```
ArnaudCharlet, line 160: it's cleaner to use arithmetic comparison rather than string comparison in this context since we know $# is a number:
if [ $# -eq 1 ]; then

pmderodat: Thanks! Should be fixed now.

ArnaudCharlet (approved):

PR #664, opened 2020-12-16
Fix memoization of resolve_own_names

raph-amiard:


extensions/analysis/implem_decls
```diff
@@ -0,0 +1,12 @@
...
+    Rebindings    : Env_Rebindings;
+    Return_Value  : Boolean;
+end record;
+
+package Nameres_Maps is new Ada.Containers.Hashed_Maps
+  (Key_Type => Bare_Ada_Node,
+   Element_Type => Resolution_Val,
+   Hash => Hash,
+   Equivalent_Keys => "=",
+   "=" => "=");
```
pmderodat, line 12: Please align arrows

extensions/src/libadalang-implementation-extensions.adb
```diff
@@ -650,4 +654,45 @@ package body Libadalang.Implementation.Extensions is
...
+      use Nameres_Maps;
+   begin
+
+      --  There was already resolution for this node, and it's the same
+      --  rebindings, and the cache key is still fresh: just return
+      --  existing result.
+
+      if Nameres_Maps.Has_Element (C)
+        and then Element (C).Cache_Version >= Node.Unit.Context.Cache_Version
+        and then Nameres_Maps.Element (C).Rebindings = E_Info.Rebindings
```
pmderodat, line 680: Just in case: can it be a problem if we reuse resolution results while the rebindings are the same, but not the node metadata?

raph-amiard: No, that's why I discriminated on rebindings, not node metadata

pmderodat (approved):

PR #679, opened 2021-01-20
Work on case insensitivity

pmderodat:


raph-amiard (approved):

raph-amiard:
I think it's "case insensitivity" ;)

PR #681, opened 2021-01-21
Fix GitHub Action for build binaries.

reznikmm:
Make `gprinstall --uninstall` optional, because the install dir is
cached, so gnatcoll/gpr could be uninstalled.

Drop extra spaces.


pmderodat (approved):

pmderodat:
Thanks for the fix, Maxim.

Could you open PRs from your fork instead of pushing branches in the upstream repository in the future? Thanks in advance!

PR #682, opened 2021-01-27
manage.py: add generate args to customize version number and build date

pmderodat:


raph-amiard:
Why is that in LAL and not in langkit ? Seems generally useful

pmderodat:
Because @senier (correct me if I’m wrong) needs the version to be included in sources (i.e. determined in the sources tracked by version control), whereas we (LAL) want it to be determined at code generation time (to be whatever the current release build/build date that comes from the environment).

senier:
> Because @senier (correct me if I’m wrong) needs the version to be included in sources (i.e. determined in the sources tracked by version control),

Thinking about it, the version could also be kept in the setup.py or Makefile which are under version control and passed to the code generation process. My feeling, though, is that it would be nicer to have the version close to the parser definition.

pmderodat:
Thanks for the feedback! Hm… would having both the CompileCtx-level settings suggested in https://github.com/AdaCore/langkit/pull/468 plus the `--version`/`--build-date` arguments available, and make the latters override (when present) the formers be okay for both of you? (so that all this logic is built in Langkit)

senier:
> Thanks for the feedback! Hm… would having both the CompileCtx-level settings suggested in [AdaCore/langkit#468](https://github.com/AdaCore/langkit/pull/468) plus the `--version`/`--build-date` arguments available, and make the latters override (when present) the formers be okay for both of you? (so that all this logic is built in Langkit)

I think this flexibility would be great! I just wonder whether the version should be replaced silently - I can't think of a use case where you set the version in the context and subsequently overwrite on the command line. Maybe the CLI should fail when the version is set already in the context?

pmderodat:
This PR has been obsoleted by AdaCore/langkit#468, so closing it.

PR #695, opened 2021-02-19
U219-050 fix memory corruption in OCaml binding for auto_provider

yakobowski:
The C strings allocated to represent the list of files in the project
were garbage-collected too aggressively by the GC, resulting in memory
corruption. Switch to a version of the code where we keep more GC roots.

Also explicitly set the last element of the array passed to LAL
to NULL. It is not clear this was done implicitly by Ctypes.

### DIFF
--- a/extensions/ocaml_api/unit_providers/module_struct
+++ b/extensions/ocaml_api/unit_providers/module_struct
@@ -57,13 +57,21 @@
 
   let create_auto_provider =
     foreign ~from:c_lib "${capi.get_name("create_auto_provider")}"
-      (ptr string @-> string @-> raisable c_type)
+      (ptr (ptr char) @-> string @-> raisable c_type)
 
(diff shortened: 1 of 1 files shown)


extensions/ocaml_api/unit_providers/module_struct
```diff
@@ -56,14 +56,22 @@
       wrap result
 
   let create_auto_provider =
-    foreign ~from:c_lib "${capi.get_name("create_auto_provider")}"
-      (ptr string @-> string @-> raisable c_type)
+    foreign ~from:c_lib "${capi.get_name("create_auto_provider")}"
```
pmderodat, line 59: Could you remove the trailing whitespaces?

pmderodat (changes requested):

pmderodat (approved):

PR #697, opened 2021-02-24
TC18-015: Fix nameres of fully qualified name for local variable.

Roldak:


ada/ast.py
```diff
@@ -11507,7 +11507,7 @@ def env_elements_baseid():
         # expressions in the baseid env_elements. We should solve that.
```
raph-amiard, line 11507: No entry file ?

raph-amiard (changes requested):

raph-amiard (approved):

PR #709, opened 2021-03-30
U324-040: Adapt lookup cache configuration in nameres

raph-amiard:
To account for corresponding changes in langkit


pmderodat (commented):
LGTM with the minor reformatting

pmderodat (approved):

PR #713, opened 2021-04-02
Make prod builds of libadalang-implementation.adb faster

pmderodat:


extensions/prod_additional_flags
```diff
@@ -2,3 +2,13 @@
 
 for Switches ("libadalang-unit_files.adb") use
    Common_Ada_Cargs & Manual_Ada_Cargs & ("-O1");
+
+--  Code for properties are just too big, which leads GCC to print:
+--
+--  libadalang-implementation.adb:[...] note: variable tracking size limit
+--     exceeded with '-fvar-tracking-assignments', retrying without
+--
+--  Disabling var-tracking from the start saves compilation time (2min50 down
```
Roldak, line 11: problem at the end of the sentence :smile:

pmderodat: Nooooo, this means I have to re-do measurements :'( Thanks for noticing this

pmderodat: Now fixed! I also fixed the patch: was using cargs for “manual” sources whereas this file is actually generated.

Roldak (approved):

PR #742, opened 2021-05-17
T226-048: Fix crash for overloaded top-level entities.

Roldak:


user_manual/changes/U601-009.yaml
```diff
@@ -0,0 +1,6 @@
+type: bugfix
```
raph-amiard, line 1: Hmmm :thinking: wrong change entry ?

raph-amiard (changes requested):

ada/ast.py
```diff
@@ -1401,6 +1408,29 @@ def env_hook_basic_decl():
...
         )
 
+    @langkit_property()
+    def populate_body_unit():
+        """
+        For library-level subprogram declarations, we also want
+        to populate the unit containing the body, so that the parent
+        package always has the same two values for this env entry of
+        this subprogram, no matter if we started by populating the
+        declaration or the body. This has to be done since the named
```
raph-amiard, line 1418: This comment won't age well :P Did you put it so that new LAL developers can come to you in a few years and ask you about the Great Named Envs Transition of 2020 ? 😁 Also I don't understand the end of the comment. And the "no matter if we started with ..." is probably misleading too since which we started with in the named envs paradigm is irrelevant.

Tell me if this summarizes the issue correctly:

>>> For library-level subprogram declarations, we always want to populate the unit containing the body, so that the lexical envs always contain the spec and the body, no matter which was initially requested

raph-amiard (changes requested):

ada/ast.py
Roldak: It's perfect! I'll update it.

Roldak: Done!

raph-amiard (approved):

PR #743, opened 2021-05-28
Move GitHub Actions to GNAT Community 2021

reznikmm:


.github/workflows/build-binaries.yml
```diff
@@ -60,8 +60,15 @@ jobs:
       - uses: actions/cache@v2
         with:
           path: ./cached_gnat
-          key: ${{ runner.os }}-gnat-ce-2020
+          key: ${{ runner.os }}-gnat-ce-2021
+      - name: Get GNAT Community 2021 toolchain
```
raph-amiard, line 64: Why do you add it here if it wasn't needed before ?

reznikmm: We have no Mac OS toolchain in gnat ce 2021. So, for Mac OS it installs gnat-ce-2020 while for other OSs gnat-ce-2021.

raph-amiard (approved):

PR #750, opened 2021-06-17
SC10-014: Add warning about the index being 0-based

raph-amiard:


ada/ast.py
```diff
@@ -4540,6 +4540,8 @@ def index_type(dim=Int):
         """
         Return the index type for dimension ``dim`` for this type, if
         applicable.
+
+        .. WARNING:: ``dim`` is 0-based, so the first index_type is at index 0.
```
pmderodat, line 4544: Double backquotes are missing around `index_type`

pmderodat (approved):

PR #757, opened 2021-06-24
U624-010: Fix support for 'Base attribute.

Roldak:


user_manual/changes/U624-010.yaml
```diff
@@ -0,0 +1,7 @@
+type: bugfix
+title: Fix support of ``'Base`` attribute
+description: |
+    This patch fixes a bug where using the ``'Base`` attribute as the prefix
```
raph-amiard, line 4: *adds support for the base attribute ? We didn't have it before

raph-amiard (approved):

user_manual/changes/U624-010.yaml
Roldak: No we had some support for it!

PR #763, opened 2021-07-19
TC28-004: Fix previous part on types and subprogram

Roldak:
Also fixes U323-012 and U421-013!


raph-amiard (changes requested):
As discussed live, change entries are wrong

raph-amiard (approved):

PR #775, opened 2021-08-03
Find references improvements

raph-amiard:


ada/ast.py
```diff
@@ -12055,12 +12058,46 @@ def find_all_references(units=AnalysisUnit.array,
...
-        refs = Var(origin.bind(Self, all_units.mapcat(lambda u: u.root.then(
-            lambda r: dn.find_refs(r.as_bare_entity)
+        # Set of all roots to look into. Here, we try to optimize the set of
+        # roots we look into, with the following heuristics:
+        all_roots = Var(Cond(
+            # If the entity we're looking for is a parameter specification,
+            # then only look in the spec & the body of the subprogram.
+            dn.basic_decl.is_a(T.ParamSpec),
+            [dn.basic_decl.semantic_parent,
+             dn.next_part.basic_decl.semantic_parent],
```
Roldak, line 12087: Unless I'm missing something, I think this is too constraining, as we lose references done through calls with named parameters with can be made from external units, e.g:
```
Foo (X => 2);
     ^ here
```

ada/ast.py
```diff
@@ -12055,12 +12058,46 @@ def find_all_references(units=AnalysisUnit.array,
...
+
+            # If we determined that the entity we're looking for is body local,
+            # then we ignore the set of units passed by the user, and only look
+            # for the entity in its own unit.
+            is_local,
+            [dn.basic_decl.parent.parents.find(lambda p: p.is_a(T.Body))],
```
Roldak, line 12093: I think separates can have visibility on entities defined in a body, so this is also too constraining.

raph-amiard:
Let's put this on the back burner

PR #782, opened 2021-08-06
U526-014: Get rid of IndexConstraint & DiscriminantConstraint

raph-amiard:
Replace them by a single CompositeConstraint, with p_is_index_constraint
and p_is_discriminant_constraint properties.

This avoids the wrong parsing of some index constraints as discriminant
constraints.

On the other hand, it means that Libadalang will happily parse wrong
code such as:

    A : My_Type (Disc => 12 .. 15);

In line with other such decisions we make about parsing a superset of
what is correct Ada.


Roldak (approved):

pmderodat:
Closing obsolete PR, changes reviewed and merged in another system.

PR #790, opened 2021-08-12
U430-027: Add DefiningName.fully_qualified_name_array

danielmercier:


ada/ast.py
```diff
@@ -12118,6 +12118,15 @@ def unique_identifying_name():
         """
         return Entity.basic_decl.unique_identifying_name_impl(Entity)
 
+    @langkit_property(public=True, return_type=T.Symbol.array)
+    def fully_qualified_name_array():
+        """
+        Return the fully qualified name corresponding to this defining name, as
+        an array of symbols.
+        """
+        return Entity.basic_decl.fully_qualified_name_impl(
```
raph-amiard, line 12127: The LAL/langkit coding style wrt. parens is either both on the same line, or closing paren on its own line, so here sth like:

```python
return Entity.basic_decl.fully_qualified_name_impl(
    dn=Entity
).map(lambda t: t.to_symbol)
```

raph-amiard (approved):

PR #793, opened 2021-08-13
U812-023: Fix Base_Subp_Decls for derived types in non-package scope.

Roldak:


ada/ast.py
```diff
@@ -5428,7 +5428,7 @@ def direct_primitive_subps():
                 # For a derived type, overriding primitives can be declared in
                 # scope even if it is not a package scope.
                 Self.type_def.is_a(DerivedTypeDef),
-                Self.parent.parent.parent.cast(DeclarativePart).then(
```
raph-amiard, line 5431: So the code was just looking at the wrong level all the time ?

Roldak: Yes -_-'

raph-amiard (approved):

PR #795, opened 2021-08-13
U813-022: Implement relation operators in expr evaluator.

Roldak:


extensions/src/libadalang-expr_eval.adb
```diff
@@ -489,6 +489,89 @@ package body Libadalang.Expr_Eval is
...
+               function Bool (X : Boolean; N : LAL.Ada_Node'Class := E)
+                  return Eval_Result renames Create_Bool_Result;
+
+               BO : constant LAL.Bin_Op := E.As_Bin_Op;
+               Op : constant LAL.Op := BO.F_Op;
+               L  : constant Eval_Result := Expr_Eval (BO.F_Left);
+               R  : constant Eval_Result := Expr_Eval (BO.F_Right);
+            begin
+               if L.Kind /= R.Kind then
+                  --  TODO??? There are actually some rules about implicit
```
raph-amiard, line 503: What are you thinking about ? More information would be welcome here, because looking at this code in 6 months you might not remember what you were referring to ;)

raph-amiard (changes requested):

extensions/src/libadalang-expr_eval.adb
Roldak: Oops that's a leftover from a copy-paste. Removed the TODO!

raph-amiard (approved):

PR #796, opened 2021-08-17
U812-028: Fix is_dispatching_call on abstract subps.

Roldak:


ada/ast.py
```diff
@@ -9064,7 +9064,7 @@ def is_dispatching_call_impl(decl=T.BasicDecl.entity):
 
         return Or(
             # A call to an abstract subprogram is necessarily dispatching
-            decl.is_a(AbstractSubpDecl, AbstractFormalSubpDecl),
```
raph-amiard, line 9067: I don't understand this fix, and even if it is correct you should update the comment above, since now this branch is only concerned with formal abstract subprograms.

But in any case, in the case of your reproducer, `decl` here should be the overriding foo declaration at `pkg-child.ads:4`, which is not an `AbstractSubpDecl`, so I don't get what's going on here

raph-amiard: Isn't this whole branch strange when confronted to RM `3.9.2` where it seems that whether a call is dispatching is entirely determined by whether the controlling expression is dynamically tagged, and not by the nature of the subprogram itself ?

Roldak: It returned a wrong result because LAL still has trouble determining the exact subprogram called when a tagged hierarchy is involved, due to elements ordering/visibility.
My test reproduces such a scenario: `decl` here is the declaration from pkg.ads and not pkg-child.ads, meaning this condition would make it instantly return True although the call is not dispatching.
The appropriate fix is indeed to fix name resolution for such calls, but in the meantime this condition (at least for `AbstractSubpDecl`) never does any good (as you noted) so I figured it could be removed, as it also has the side benefit of fixing this issue. As for `AbstractFormalSubpDecl`, I think those cause some trouble internally as our routines handle primitive operations only but abstract formal subprograms are not primitive of any type (see ramification in 3.9.3).
I'll update the comment!

raph-amiard: Ok! So it means that if you run `nameres` on your example, it would actually return wrong results right ?

We should keep track of this somewhere at least, be it a TN or an xfail'ed test ...

raph-amiard (approved):

ada/ast.py
Roldak: Yes that's it! And this problem (at least one occurrence of it) is tracked under U609-030. Thanks for the review!

PR #798, opened 2021-08-17
U816-006: Fix parser for pragma argument assocs.

Roldak:


user_manual/changes/U816-006.yaml
```diff
@@ -0,0 +1,9 @@
+type: api-change
+title: Fix parser for pragma argument assocs
```
raph-amiard, line 2: I think it should be clearer that this breaks `PragmaArgumentAssoc` in the title, since this is an entry for an api-change, we care less about the fix than about the API change.

raph-amiard (changes requested):

raph-amiard (approved):

Roldak:
Fixed by PR #800

PR #801, opened 2021-08-20
U818-018: Stabilize behavior of ``P_Next_Part_For_Decl``

Roldak:


user_manual/changes/U818-018.yaml
```diff
@@ -0,0 +1,8 @@
+type: bugfix
+title: Stabilize behavior of ``P_Next_Part_For_Decl``
+description: |
+    This change improves the stability of ``P_Next_Part_For_Decl`` to be
```
raph-amiard, line 4: I'm not sure stability is the right word. Consistency/Coherency seems better here (same in the commit message and change entry title).

Unstability in my mind would let one think that this property would return different results for the same inputs sometimes, which is definitely not something you want to even imply ;)

raph-amiard (changes requested):

user_manual/changes/U818-018.yaml
Roldak: Done!

PR #809, opened 2021-09-03
U902-020: Add is_ghost_code property

raph-amiard:


ada/ast.py
```diff
@@ -7242,6 +7252,15 @@ class Pragma(AdaNode):
 
     xref_entry_point = Property(True)
 
+    @langkit_property(public=True)
+    def is_ghost_code():
+        """
+        Return whether this declaration is ghost code or not. See SPARK RM 6.9.
+        """
+        return Entity.id.sym.any_of(
+            'Assert', 'Assume'
```
Roldak, line 7261: Missing `Loop_Invariant`, maybe also `Invariant` (for types), `Pre`, `Post`? (and their 'Class variant?)

ada/ast.py
```diff
@@ -15287,6 +15306,31 @@ class Stmt(AdaNode):
 
     xref_entry_point = Property(True)
 
+    @langkit_property(public=True)
+    def is_ghost_code():
+        """
+        Return whether this declaration is ghost code or not. See SPARK RM 6.9.
```
Roldak, line 15312: copy-pasta: `declaration` -> `statement`

Roldak (changes requested):
What about Ada 2012 aspects, shouldn't we also provide `p_is_ghost_code` on them like we have for pragmas? (If so, probably `AdaNode.is_contract_aspect` can help)

raph-amiard:
> What about Ada 2012 aspects, shouldn't we also provide `p_is_ghost_code` on them like we have for pragmas? (If so, probably `AdaNode.is_contract_aspect` can help)

It's not necessary for gnatcov because gnatcov treats aspects differently (since they're never in lists of decls/stmts), hence why I didn't add them

Roldak (approved):

PR #816, opened 2021-09-09
U909-006: Fix is_static_expr for prefix of attribute refs.

Roldak:


thvnx (approved):
Not an expert... yet :D, but LGTM.

PR #832, opened 2021-09-24
T320-010: Adapt to langkit changes.

Roldak:


ada/ast.py
```diff
@@ -7272,11 +7273,13 @@ def env_names():
...
-        add_to_env_by_name(
+        add_to_env_kv(
             key=Entity.name_symbol,
             val=Self,
-            name=Self.child_decl_initial_env_name(False),
-            fallback_env=Self.children_env
+            dest_env=Self.child_decl_initial_env_name(False).then(
+                lambda name: named_dest_env(name),
+                default_val=current_dest_env()
+            ),
```
pmderodat, line 7282: It seems that this is now the same expression over and over: what about factoring this into a property that returns a mapping (`T.env_assoc`), and just have `add_to_env(Self.<this-property>)` everywhere?

(or two properties, as we seem to have another series of duplicates at the end of the patch)

Roldak: Good catch, done!

ada/ast.py
```diff
@@ -1664,6 +1657,37 @@ def child_decl_initial_env_name(private_part=(T.Bool, False)):
...
+        return Self.child_decl_initial_env_name(private_part).then(
+            lambda name: named_env(name),
+            default_val=direct_env(Self.default_initial_env)
+        )
+
+    @langkit_property(return_type=T.env_assoc)
+    def child_decl_env_assoc():
+        """
+        Return the env association that describes where to register this
+        basic declaration. For child declaration in particular, this orders
```
pmderodat, line 1679: For child declarations* (or “For a child declaration”)

ada/ast.py
```diff
@@ -2908,6 +2944,31 @@ def previous_part_env_name():
             Self.top_level_env_name.to_symbol
         )
 
+    @langkit_property(return_type=T.env_assoc)
+    def previous_part_link_env_assoc():
+        """
+        Return the env association that describes where to add a `__nextpart`
+        entry for this body, if it corresponds to a non-overloadable entity
+        (i.e not a subprogram).
```
pmderodat, line 2952: i.e.* (missing dot)

pmderodat: Btw, this also applies to entries, right? Do we consider them as subprograms?

Roldak: Thanks for raising this, I will add a small comment to say this isn't used by entries. In short: entry navigation is not handled the same way because they can never appear as a top level entity and are always contained in a declaration (task or protected object), which itself relies on that mechanism.

Roldak: Done!

ada/ast.py
Roldak: Thanks!

ada/ast.py
pmderodat: Ok thanks!

ada/ast.py
```diff
@@ -3298,6 +3363,16 @@ def env_names():
...
         ])
 
+    @langkit_property()
+    def previous_part_link_env_assoc():
+        return new_env_assoc(
+            key='__nextpart',
+            val=Self,
+            dest_env=Self.top_level_env_name.to_symbol.then(
+                lambda name: named_env(name)
+            )
```
pmderodat, line 3373: Couldn’t this be simplified to `named_env(Self.top_level_env_name.to_symbol)`?

Roldak: Yes! done

pmderodat (approved):

PR #838, opened 2021-09-28
U623-025 Fix bad highlighting in generated HTML

yannickmoy:
Remove 's' which breaks highlighting after a double-quoted string.
Fix English typo.


pmderodat (approved):

pmderodat:
Thanks Yannick!

PR #851, opened 2021-10-14
TB01-005: Name resolution of `limited with` clauses.

Roldak:
The previous implementation failed to resolve names in a limited with clause
designating a child unit of the current unit, because the behavior of
`xref_no_overloading` takes into account visibility as seen from node at
the origin of the query: since a with clause appears before the declaration
of the current unit, an env query of any name referencing that unit or one of
its children would return an empty result.

### DIFF
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -6890,9 +6890,24 @@ class WithClause(AdaNode):
     has_private = Field(type=Private)
     packages = Field(type=T.Name.list)
 
+    @langkit_property(return_type=Equation)
+    def child_unit_xref_equation(name=T.Name):
+        """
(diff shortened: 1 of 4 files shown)


ada/ast.py
```diff
@@ -6890,9 +6890,19 @@ class WithClause(AdaNode):
     has_private = Field(type=Private)
     packages = Field(type=T.Name.list)
 
+    @langkit_property(return_type=Equation)
+    def child_unit_xref_equation(name=T.Name):
```
raph-amiard, line 6894: Please comment/add a docstring here

Roldak: Done!

PR #853, opened 2021-10-18
TA29-030: Fix introduced null deref in last commit.

Roldak:


pmderodat (changes requested):
Actually could you add a testcase for this?

PR #868, opened 2021-11-17
Add `EntryDecl.p_accept_stmts` property

thvnx:
This new property returns an array of accept statements calling
the entry declaration.

TN: S916-015

### DIFF
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -13962,6 +13962,42 @@ class EntryDecl(BasicSubpDecl):
         """
         return Entity.spec.family_type.cast(TypeExpr)._.designated_type
 
+    @langkit_property(public=True, return_type=T.AcceptStmt.entity.array)
+    def accept_stmts():
+        """
+        Return an array of accept statements corresponding to this entry.
+        """
+        return Entity.find_accept_stmts(
+            # Find the body part of the task declaration for the current entry
+            Entity.parent_basic_decl.match(
+                lambda st=T.SingleTaskTypeDecl.entity: st.parent_basic_decl,
+                lambda tt=T.TaskTypeDecl: tt,
+                lambda _: PropertyError(
+                    T.BasicDecl.entity,
+                    "Can only be called on EntryDecl in Tasks"
+                )
+            ).body_part_for_decl
+        )
+
+    @langkit_property(return_type=T.AcceptStmt.entity.array)
+    def find_accept_stmts(root=T.AdaNode.entity):
+        """
+        Find all accept statements in all children of ``root`` that correspond
+        to this entry declaration.
+        """
+        return root.children.then(
+            lambda child: child.filter(lambda n: Not(n.is_null))
+            .mapcat(lambda n: Entity.find_accept_stmts(n))
+        ).concat(
+            root.cast(T.AcceptStmt).then(
+                lambda stmt: If(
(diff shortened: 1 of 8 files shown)


ada/ast.py
```diff
@@ -13962,6 +13962,34 @@ def family_type():
...
+        """
+        Find all accept statements in all children of ``root`` that call this
+        entry declaration.
+        """
+        return root.children.then(
+            lambda child: child.filter(lambda n: Not(n.is_null))
+            .mapcat(lambda n: Entity.find_accept_stmts(n))
+        ).concat(
+            root.cast(T.AcceptStmt).then(
+                lambda stmt: stmt.singleton.filter(
```
Roldak, line 13986: Minor:
`.singleton.filter` is a bit awkward IMO, I'd rather have:
```
If(stmt.corresponding_entry == Entity,
   stmt.singleton,
   No(AcceptStmt.entity.array))
```

ada/ast.py
```diff
@@ -13962,6 +13962,34 @@ def family_type():
         """
         return Entity.spec.family_type.cast(TypeExpr)._.designated_type
 
+    @langkit_property(public=True)
```
Roldak, line 13965: Personally I prefer specifying the return type for all properties, especially public ones (avoid changing the API by mistake).

Roldak (changes requested):
Only minor comments, looks good otherwise!

user_manual/changes/S916-015.yaml
```diff
@@ -0,0 +1,6 @@
+type: new-feature
+title: Add property ``EntryDecl.p_accept_stmts``
+description: |
+    This new property returns an array of accept statements calling
```
Roldak, line 4: `calling` is not really the right word here. What about `When invoked on a task entry declaration, this property returns the array of all its corresponding accept statements.`?

Roldak (changes requested):

Roldak (changes requested):
Could you also test the property on a TaskTypeDecl's entry?

user_manual/changes/S916-015.yaml
thvnx: Sounds good, thanks.

thvnx:
> Could you also test the property on a TaskTypeDecl's entry?

Done. Well, it was useful to do so as it revealed a bug in the code. TaskTypeDecl are now correctly supported.

ada/ast.py
```diff
@@ -13962,6 +13962,38 @@ def family_type():
...
+    @langkit_property(public=True, return_type=T.AcceptStmt.entity.array)
+    def accept_stmts():
+        """
+        Return an array of accept statements corresponding to this entry.
+        """
+        return Entity.find_accept_stmts(
+            # Find the body part of the task declaration for the current entry
+            Entity.parent_basic_decl.match(
+                lambda st=T.SingleTaskTypeDecl.entity: st.parent_basic_decl,
+                lambda other: other
```
Roldak, line 13974: Since you already have that, what about adding an explicit case for `TaskTypeDecl` and raising a PropertyError for any other case?

Roldak (changes requested):

ada/ast.py
thvnx: Sure, done!

Roldak (approved):

PR #894, opened 2021-12-02
Add Ada 2022 iterator_filter support for iterator specification

thvnx:
This patch implements the iterator_filter feature introduced in Ada
2022, AI12-0250.

TN: UB16-009

### DIFF
--- a/testsuite/tests/name_resolution/iterated_component_association/iterassoc_filter.adb
+++ b/testsuite/tests/name_resolution/iterated_component_association/iterassoc_filter.adb
@@ -1,0 +1,26 @@
+procedure Iterassoc_Filter is
+   type R is range 1 .. 5;
+   type A is array (R) of Float;
+
+   X : A := (1.0, 2.0, 3.0, 4.0, 5.0);
+
+   M : A := (for I in R when X (I) >= 3.0 => X (I) * X (I));
+   pragma Test_Statement;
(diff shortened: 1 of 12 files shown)


testsuite/tests/name_resolution/iterated_component_association/iterassoc_filter.adb
```diff
@@ -0,0 +1,26 @@
...
+
+   M : A := (for I in R when X (I) >= 3.0 => X (I) * X (I));
+   pragma Test_Statement;
+
+   O : A := (for F : Float of X when F >= 3.0 => F * F);
+   pragma Test_Statement;
+
+   P : A := (for F of X when F >= 3.0 => F * F);
+   pragma Test_Statement;
+
```
Roldak, line 15: Iterator filters are also valid in quantified expressions, could you add a little test case here? (It should work already with what you've done)

Roldak (changes requested):

Roldak (approved):

testsuite/tests/name_resolution/iterated_component_association/iterassoc_filter.adb
thvnx: As discussed live, those tests will be added later since they require additional development first.

PR #895, opened 2021-12-02
UC02-034: Fix `P_Next_Part_For_Decl` on generic declarations.

Roldak:


testsuite/tests/navigation/generic_pkg_top_level_2/foo.adb
```diff
@@ -0,0 +1,3 @@
+package body Foo is
```
thvnx, line 1: No test directives?

testsuite/tests/navigation/generic_pkg_top_level_2/foo.ads
```diff
@@ -0,0 +1,5 @@
+generic
```
thvnx, line 1: Same

testsuite/tests/navigation/generic_pkg_top_level_2/foo.adb
thvnx: Nevermind :D, that's good actually.

Roldak: This is only to test the `p_body_part_for_decl` property, so the directive is in the `test.adb` file.

thvnx (approved):
LGTM

PR #896, opened 2021-12-02
Add Ada 2022 `'Index` attribute nameres support

thvnx:
TN: UC02-026

### DIFF
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -15218,6 +15218,9 @@ class AttributeRef(Name):
             rel_name == 'Put_Image',
             Entity.put_image_equation,
 
+            rel_name == 'Index',
+            Entity.index_equation,
+
             PropertyError(Equation, "Unhandled attribute")
         )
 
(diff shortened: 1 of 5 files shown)


ada/ast.py
```diff
@@ -15798,6 +15801,21 @@ def put_image_equation():
                  eq_prop=BaseTypeDecl.matching_formal_type)
         )
 
+    @langkit_property(return_type=Equation, dynamic_vars=[env, origin])
+    def index_equation():
+        """
+        Return the xref equation for the `Index` attribute.
+        """
+        typ = Var(
```
Roldak, line 15809: Unfortunately this won't handle the following case:
```
entry E (Entry_Id_Mod);
pragma Precondition (E'Index < 2);
```
You should instead use an approach based on lexical environments.

Roldak (changes requested):

Roldak (approved):

PR #899, opened 2021-12-08
Add `is_constant_object` property

thvnx:
TN: U330-040

### DIFF
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -2995,6 +2995,14 @@ class BasicDecl(AdaNode):
             default_val=Entity
         )
 
+    @langkit_property(kind=AbstractKind.abstract_runtime_check,
+                      public=True, return_type=Bool)
+    def is_constant_object():
+        """
+        Return whether this object is constant or not.
+        """
+        pass
+
 
 class ErrorDecl(BasicDecl):
     """
@@ -7755,6 +7763,10 @@ class ParamSpec(BaseFormalParamDecl):
 
     type_expression = Property(Entity.type_expr)
 
+    @langkit_property(return_type=Bool)
+    def is_constant_object():
+        return Self.mode.is_a(Mode.alt_in, Mode.alt_default)
+
     @langkit_property()
     def defining_env():
         return Entity.type_expr.defining_env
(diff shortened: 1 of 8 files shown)


ada/ast.py
```diff
@@ -8554,6 +8561,33 @@ class ObjectDecl(BasicDecl):
...
+
+            # Renaming clause is:
+            Entity.renaming_clause.then(
+                lambda rc: If(
+                    # Always constant if it renames a function_call
+                    rc.renamed_object.is_call,
+                    True,
+                    # Constant if the renamed object is constant
+                    rc.renamed_object.referenced_decl
+                    .cast_or_raise(T.ObjectDecl).is_constant_object
```
Roldak, line 8586: The renamed object could be a parameter as well. I think you should have an abstract `is_constant_object` property on `BasicDecl` so that you can recurse easily here (it will also be easier to use for users).

ada/ast.py
```diff
@@ -8554,6 +8561,33 @@ class ObjectDecl(BasicDecl):
...
 
+    @langkit_property(public=True, return_type=Bool)
+    def is_constant_object():
+        """
+        Return whether this object is constant or not.
+        """
+        return Or(
+            Self.has_constant.as_bool,
+
+            # A GenericFornalObjDecl is constant if the Mode is `in`
```
Roldak, line 8572: Typo: `GenericFormalObjDecl`

Roldak (changes requested):

ada/ast.py
```diff
@@ -7755,6 +7765,13 @@ class ParamSpec(BaseFormalParamDecl):
 
     type_expression = Property(Entity.type_expr)
 
+    @langkit_property(public=True, return_type=Bool)
+    def is_constant_object():
+        """
```
raph-amiard, line 7770: You only need this docstring on the base property (and same for the public annotation). Technically that's also true for the return type but you can leave it - especially since it will be mandatory in lkt

ada/ast.py
```diff
@@ -2995,6 +2995,16 @@ def wrap_public_reference():
             default_val=Entity
         )
 
+    @langkit_property(public=True, return_type=Bool)
```
raph-amiard, line 2998: Use abstract_runtime_check to have this behavior automatically (check other properties in ast.py)

raph-amiard (dismissed):

ada/ast.py
thvnx: Should I also modify the propertyError message for something like `should not happen` instead?

thvnx: Even better, as mentioned live by @Roldak , just use `pass` here.

Roldak (approved):

Roldak (approved):

PR #900, opened 2021-12-08
Fix `EntrySpec` xref_equation

thvnx:
This change fixes the xref_equation for `EntrySpec` name resolution.

TN: UC02-047

### DIFF
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -14695,6 +14695,11 @@ class EntrySpec(BaseSubpSpec):
     )
     returns = Property(No(T.TypeExpr.entity))
 
+    xref_equation = Property(Entity.family_type.then(
+        lambda r: r.sub_equation,
+        default_val=LogicTrue()
+    ))
+
 
 class Quantifier(AdaNode):
     """
--- a/user_manual/changes/UC02-047.yaml
+++ b/user_manual/changes/UC02-047.yaml
@@ -1,0 +1,6 @@
+type: bugfix
+title: Resolve references to entry families.
+description: |
+    This change fixes a bug where references to entry families were missing from
+    name resolution.
+date: 2021-12-08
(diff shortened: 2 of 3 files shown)


user_manual/changes/UC02-047.yaml
```diff
@@ -0,0 +1,5 @@
+type: bugfix
+title: Fix `EntrySpec` xref_equation.
+description: |
+    This change fixes the xref_equation for `EntrySpec` name resolution.
```
Roldak, line 4: These change entries do not target LAL developers but LAL users, and they have no clue what you are talking about here :D

ada/ast.py
```diff
@@ -14695,6 +14695,11 @@ class EntrySpec(BaseSubpSpec):
     )
     returns = Property(No(T.TypeExpr.entity))
 
+    xref_equation = Property(Entity.family_type.then(
+        lambda r: r.sub_equation,
+        default_val=LogicTrue())
+    )
```
Roldak, line 14701: The closing parenthesis for `then(` should be on line 14701 as well since it was opened on the same line as the one for `Property(`.

Roldak (changes requested):

user_manual/changes/UC02-047.yaml
thvnx: Should I remove it or move it somewhere else?

Roldak: No you should rewrite it with a different wording, such that users can understand it (they don't know what xref_equation is). From their POV, what didn't work before but will work now thanks to your patch?

thvnx: Sure, sorry about that.

thvnx: Hope it's better now?

Roldak (approved):

PR #901, opened 2021-12-09
Make subtype_mark optional in object renames

thvnx:
UC08-034

### DIFF
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -8716,6 +8716,18 @@ class ExtendedReturnStmtObjectDecl(ObjectDecl):
     pass
 
 
+class NoTypeObjectRenamingDecl(ObjectDecl):
+    """
+    Object declaration without subtype indication. This node has been
+    introduced to cover a special case for `ObjectDecl`, where `type_expr`
+    is made optional (AI12-0275), and therefore cannot fit in an
+    `ObjectDecl`.
+    """
+    @langkit_property()
+    def xref_equation():
+        return Entity.renaming_clause.renamed_object.sub_equation
+
+
 class DeclarativePart(AdaNode):
     """
     List of declarations.
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -8543,6 +8543,8 @@ class NumberDecl(BasicDecl):
 
     xref_equation = Property(Entity.expr.sub_equation)
 
+    is_constant_object = Property(True)
+
 
 class ObjectDecl(BasicDecl):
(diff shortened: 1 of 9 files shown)


user_manual/changes/UC08-034.yaml
```diff
@@ -0,0 +1,6 @@
+type: new-feature
+title: Make subtype_mark optional in object renames
+description: |
+    The subtype_mark in object renaming declarations is made optional as per
+    AI12-0275 (Ada 2022 feature).
```
raph-amiard, line 5: This is not sufficient: You're introducing a new node type in LAL, this needs to be documented here as well, since it's not merely as simple as making a field optional.

ada/ast.py
```diff
@@ -8716,6 +8719,16 @@ class ExtendedReturnStmtObjectDecl(ObjectDecl):
     pass
 
 
+class NoTypeRenameObjectDecl(ObjectDecl):
+    """
+    Object declaration that is part of an iterator specification without
```
raph-amiard, line 8724: Plz fix the docstring ;)

ada/ast.py
```diff
@@ -8716,6 +8719,16 @@ class ExtendedReturnStmtObjectDecl(ObjectDecl):
     pass
 
 
+class NoTypeRenameObjectDecl(ObjectDecl):
```
raph-amiard, line 8722: Let's call it `NoTypeObjectRenamingDecl`

ada/ast.py
raph-amiard: Also please explain in this docstring why we had to introduce a new node type for this (at a high level), rather than just reuse the old one.

raph-amiard (changes requested):

ada/ast.py
```diff
@@ -8716,6 +8719,18 @@ class ExtendedReturnStmtObjectDecl(ObjectDecl):
     pass
 
 
+class NoTypeObjectRenamingDecl(ObjectDecl):
+    """
+    Object declaration without subtype indication. This node has been
+    introduced to cover a special case for `ObjectDecl`, where `type_expr`
+    is made optional (AI12-0275), and therefore cannot fit in an
+    `ObjectDecl`.
```
raph-amiard, line 8727: This is RST so you need to use double backticks

raph-amiard (changes requested):

raph-amiard (approved):

PR #904, opened 2021-12-14
Improve `is_constant_object` property support

thvnx:
TN: U330-040

### DIFF
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -16622,6 +16622,8 @@ class ExceptionHandler(BasicDecl):
 
     xref_entry_point = Property(True)
 
+    is_constant_object = Property(True)
+
 
 @abstract
 class Stmt(AdaNode):
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -15050,6 +15050,8 @@ class QualExpr(Name):
 
     relative_name = Property(Entity.prefix.relative_name)
 
+    is_constant = Property(True)
+
     @langkit_property(return_type=Equation, dynamic_vars=[env, origin])
     def general_xref_equation(root=(T.Name, No(T.Name))):
         return And(
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -4257,6 +4257,8 @@ class ComponentDecl(BaseFormalParamDecl):
 
     type_expression = Property(Self.component_def.type_expr.as_entity)
 
+    is_constant_object = Property(False)
+
     @langkit_property(return_type=Equation)
     def constrain_prefix(prefix=T.Expr):
         # Simple type equivalence
@@ -14026,6 +14028,10 @@ class Identifier(BaseId):
         )
     )
 
+    @langkit_property()
+    def is_constant():
+        return Entity.referenced_decl.is_constant_object
+
 
 class StringLiteral(BaseId):
     """
@@ -14825,6 +14831,19 @@ class ForLoopVarDecl(BasicDecl):
         )
 
     env_spec = EnvSpec(add_to_env_kv(Self.name_symbol, Self))
+
+    @langkit_property()
+    def is_constant_object():
+        # TODO: add support for Constant/Variable_Indexing
+        loop_spec = Var(Entity.parent.cast(ForLoopSpec))
+        return If(
+            # IterType.alt_of loops are constant if the iterable object is
+            # constant.
+            loop_spec.loop_type.is_a(IterType.alt_of),
+            loop_spec.iter_expr.cast_or_raise(Name).is_constant,
+            # IterType.alt_in loops are always constant
+            True
+        )
 
 
 class ForLoopSpec(LoopSpec):
@@ -16037,6 +16056,15 @@ class DottedName(Name):
                 & e.cast(BasicDecl.entity).constrain_prefix(Self.prefix)
                 & Self.type_bind_var(Self.type_var, Self.suffix.type_var)
             ))
+        )
+
+    @langkit_property()
+    def is_constant():
+        # A dotted name references a constant object if the prefix or the
+        # suffix does.
+        return Or(
+            Entity.prefix.is_constant,
+            Entity.suffix.is_constant
         )
 
 
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -11850,7 +11850,7 @@ class Name(Expr):
         return Self.sym_join(Self.as_symbol_array, String(".")).to_symbol
 
     @langkit_property(kind=AbstractKind.abstract_runtime_check,
-                      return_type=Bool)
+                      public=True, return_type=Bool)
     def is_constant():
         """
         Return whether this name denotes a constant value.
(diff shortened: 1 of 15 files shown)


ada/ast.py
```diff
@@ -11839,6 +11859,13 @@ def canonical_text():
         """
         return Self.sym_join(Self.as_symbol_array, String(".")).to_symbol
 
+    @langkit_property(return_type=Bool)
+    def is_constant_reference():
```
raph-amiard, line 11863: I'd advise naming this just `is_constant`

ada/ast.py
```diff
@@ -11839,6 +11859,13 @@ def canonical_text():
...
+        """
+        Return whether the referenced declaration is constant or not.
```
raph-amiard, line 11865: "Return whether this name denotes a constant value"

ada/ast.py
```diff
@@ -14800,6 +14827,16 @@ def expr_type():
 
     env_spec = EnvSpec(add_to_env_kv(Self.name_symbol, Self))
 
+    @langkit_property()
+    def is_constant_object():
+        loop_spec = Var(Entity.parent.cast(ForLoopSpec))
+        return If(
+            loop_spec.loop_type.is_a(IterType.alt_of),
+            loop_spec.iter_expr.cast_or_raise(Name).is_constant_reference,
+            # IterType.alt_in are always constant
```
raph-amiard, line 14836: You should probably put a larger comment explaining the different cases that we went over live the other day

raph-amiard (changes requested):

ada/ast.py
```diff
@@ -14000,6 +14051,25 @@ class Identifier(BaseId):
...
+    @langkit_property()
+    def is_constant():
+        return Or(
+            Entity.referenced_decl.is_constant_object,
+
+            # An instance of a protected unit is constant within
+            # a function body of the corresponding protected type.
+            Entity.parents.find(
+                lambda n: n.cast(T.SubpBody)._.subp_spec
+                .subp_kind.is_a(SubpKind.alt_function)
```
raph-amiard, line 14063: You're not checking that the function is defined inside a protected. I imagine that still works, because if you have visibility  on those protected variables, you are necessarily in a protected object, but if that was your reasoning this should be explained in a comment IMO

ada/ast.py
```diff
@@ -14000,6 +14051,25 @@ class Identifier(BaseId):
...
 
+    @langkit_property()
+    def is_constant():
+        return Or(
+            Entity.referenced_decl.is_constant_object,
+
+            # An instance of a protected unit is constant within
+            # a function body of the corresponding protected type.
+            Entity.parents.find(
+                lambda n: n.cast(T.SubpBody)._.subp_spec
```
raph-amiard, line 14062: Wrong node type: Use `T.BaseSubpBody` to include expression functions too

ada/ast.py
```diff
@@ -14000,6 +14051,25 @@ class Identifier(BaseId):
         )
     )
 
+    @langkit_property()
+    def is_constant():
+        return Or(
+            Entity.referenced_decl.is_constant_object,
+
+            # An instance of a protected unit is constant within
```
raph-amiard, line 14059: The comment speaks about the protected unit, but here you're not checking the unit but the variables defined inside the unit, so the comment should be adjusted to talk about what you're really doing

ada/ast.py
```diff
@@ -14800,6 +14870,24 @@ def expr_type():
 
     env_spec = EnvSpec(add_to_env_kv(Self.name_symbol, Self))
 
+    @langkit_property()
+    def is_constant_object():
+        loop_spec = Var(Entity.parent.cast(ForLoopSpec))
+        return If(
+            # IterType.alt_of loops are constant if the iterable object is
+            # constant. Two cases are supported so far:
+            # 1. The iterable object is defined as constant such like a
```
raph-amiard, line 14879: * such as

ada/ast.py
```diff
@@ -14800,6 +14870,24 @@ def expr_type():
...
+            #    constant array.
+            # 2. The iterable object is not directly declared as constant but
```
raph-amiard, line 14881: I think that now that you use `is_constant`, this comment is useless

raph-amiard: Also I think you're lacking the `Constant/Variable_Dereference` part of the RM, so that should be mentioned in a comment here

raph-amiard (changes requested):

ada/ast.py
```diff
@@ -11896,6 +11914,27 @@ class CallExpr(Name):
...
+        return Cond(
+            # Case for type conversion: CallExpr has one
+            # argument and its name denotes a type declaration.
+            And(
+                Entity.params.length == 1,
+                Entity.name.referenced_decl.is_a(T.TypeDecl)
+            ),
+            Entity.params.at(0).expr.match(
+                # View conversion: constant if the object is constant
+                lambda n=Identifier: n.is_constant,
```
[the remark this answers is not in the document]
thvnx: Actually, I tried with `Name` first but it matched all the cases I could think of. Can it be something else than a Name?

For example, `1.6` in `Integer (1.6)` is a `RealLiteral`, which is a `Name`, but a value conversion (always constant).

ada/ast.py
thvnx: I'm not sure such cases are missing actually since `is_constant` property is implemented for `ExplicitDeref` names. Are you thinking of something else?

ada/ast.py
```diff
@@ -8570,14 +8574,23 @@ def is_constant_object():
...
                     True,
+
+                    # Constant if the renamed object is an ExplicitDeref of an
+                    # access constant type.
+                    rc.renamed_object.is_a(T.ExplicitDeref),
+                    rc.renamed_object.is_constant,
+
                     # Constant if the renamed object is constant
                     rc.renamed_object.referenced_decl.is_constant_object
                 )
```
Roldak, line 8589: Can't you just reuse the `Name.is_constant` property here on `rc.renamed_object`? It seems to me that this code is re-implementing some specific cases (but I may be wrong)

Roldak (changes requested):

raph-amiard (approved):

ada/ast.py
thvnx: This code was written before the `is_constant` property, explaining the duplicate implementation here. I fixed it, good catch!

ada/ast.py
```diff
@@ -12928,6 +12958,16 @@ def general_xref_equation(root=(T.Name, No(T.Name))):
             )
         )
 
+    @langkit_property()
+    def is_constant():
+        # The dereference expression is constant if its access type is
+        # constant.
+        return origin.bind(
+            Self,
+            Entity.referenced_decl.expr_type._.access_def
```
Roldak, line 12967: It's better to use `expression_type` on the `prefix` than to call `referenced_decl.expr_type`. The latter will not handle all possible cases, as the prefix of an ExplicitDeref is not necessarily a name on which you can use referenced_decl.

Roldak (changes requested):

Roldak (approved):

PR #908, opened 2022-01-04
Ada 2022 defaults for generic formal types support

thvnx:
This patch adds support for default subtypes to be specified in
generic formal type parameters.

See also:
http://www.ada-auth.org/cgi-bin/cvsweb.cgi/ai12s/ai12-0205-1.txt.

TN: V104-012

### DIFF
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -4694,6 +4694,7 @@ class BaseTypeDecl(BasicDecl):
                 has_not_null=T.NotNullAbsent.new(),
                 type_decl=Self
             ),
+            default_type=No(T.Name),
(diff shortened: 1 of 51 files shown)


raph-amiard (approved):
LGTM!

raph-amiard (changes requested):
On second thought: I think it would be possible to add the default type on the wrapper type (`GenericFormalTypeDecl`) rather than on `TypeDecl`, and it would be much better, since it doesn't make sense to have it on non generic formal type decls.

thvnx:
As discussed live, this changes can't be easily implemented at the moment. Switch to draft.

PR #910, opened 2022-01-07
V106-022: Add missing override of base_interfaces on BaseSubtypeDecl.

Roldak:


ada/ast.py
```diff
@@ -5213,8 +5213,24 @@ def is_derived_type(other_type=T.BaseTypeDecl.entity):
...
+            Let(
+                lambda
+                entity_cw=entity_can.cast(ClasswideTypeDecl),
+                other_cw=other_can.cast(ClasswideTypeDecl):
+
+                Cond(
+                    Not(entity_cw.is_null) & Not(other_cw.is_null),
+                    entity_cw.typedecl.is_derived_type(other_cw.typedecl),
+
+                    Not(entity_cw.is_null),
```
raph-amiard, line 5225: I'm not sure I understand this code, which means it needs comments :^)

raph-amiard (changes requested):

ada/ast.py
Roldak: Nice comment! I re-read this code and thought of a way to improve it but it didn't work... Turns out there was a deeper problem which I didn't see originally! The new patch doesn't touch `is_derived_type` at all.

raph-amiard (approved):

PR #911, opened 2022-01-07
V106-026: Resolve message argument of `Assert` pragmas.

Roldak:


thvnx (commented):
No change entry?

Roldak:
@thvnx This one might deserve one indeed. Added!

raph-amiard (approved):

PR #913, opened 2022-01-07
Fix ``is_constant_object`` property

thvnx:
This fix adds a missing case for the ``is_constant_object`` property
where a call expression refers to a subprogram without specification.

TN: V107-027

### DIFF
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -16414,6 +16414,8 @@ class BaseSubpBody(Body):
 
     defining_names = Property(Entity.subp_spec.name.singleton)
 
+    is_constant_object = Property(True)
+
     @langkit_property(return_type=LexicalEnv, dynamic_vars=[origin])
     def defining_env():
         return If(
--- a/testsuite/tests/properties/is_constant_object/pack.adb
+++ b/testsuite/tests/properties/is_constant_object/pack.adb
@@ -1,0 +1,9 @@
+package body Pack is
+   function My_Fun (I : Integer) return Integer is
+   begin
+      return I + 1;
+   end;
+
+   My_2 : Integer renames My_Fun (1);
+   --% node.p_is_constant_object
+end Pack;
--- a/testsuite/tests/properties/is_constant_object/pack.ads
+++ b/testsuite/tests/properties/is_constant_object/pack.ads
@@ -1,0 +1,2 @@
+package Pack is
+end Pack;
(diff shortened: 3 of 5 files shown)


testsuite/tests/properties/is_constant_object/pack.adb
```diff
@@ -0,0 +1,9 @@
+package body Pack is
+   function My_Fun (I : Integer) return Integer is
+   begin
+      return I + 1;
+   end;
+
+   My_2 : Integer renames My_Fun (1);
```
Roldak, line 7: I don't understand how this actually tests the change you made in `CallExpr.is_constant`. Surely `My_Fun` properly resolves to a declaration in this case as well (a SubpBody to be more accurate) so `referenced_decl` should not be null.

Roldak (changes requested):

testsuite/tests/properties/is_constant_object/pack.adb
thvnx: Here, `is_constant_object` is called on the `ObjectDecl` `My_2`, which, since it is a renames, ends to call `is_constant` on the CallExpr  name`My_Fun (1)` (calling `referenced_decl` on it returns null).

thvnx: As discussed live, looks like I missed something, I'm on it!

user_manual/changes/V107-027.yaml
```diff
@@ -0,0 +1,6 @@
+type: bugfix
+title: Fix ``is_constant_object`` property
+description: |
+    This fix adds a missing case for the ``is_constant_object`` property where a
+    call expression refers to a subprogram without specification.
+date: 2022-01-07
```
Roldak, line 6: A change entry is not necessary: we didn't have any release since this property was added, so no customer could be able to witness this bug.

Roldak (changes requested):

ada/ast.py
```diff
@@ -16618,6 +16618,8 @@ class SubpBody(BaseSubpBody):
 
     declarative_parts = Property(Entity.decls.singleton)
 
+    is_constant_object = Property(True)
```
Roldak, line 16621: This should be on `BaseSubpBody` IMO, otherwise it won't handle `ExprFunction`

Roldak (changes requested):

testsuite/tests/properties/is_constant_object/pack.adb
Roldak: Alright it looks better now!

Roldak (dismissed):

raph-amiard (approved):

PR #914, opened 2022-01-07
Add `Stable_Properties` aspect support

thvnx:
This change adds support for the `Stable_Properties` aspect introduced
by Ada 2022 standard revision.

### DIFF
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -179,6 +179,7 @@ class AdaNode(ASTNode):
             "Default_Initial_Condition",
             "Contract_Cases", "Test_Case",
             "Global", "Refined_Global", "Refined_State",
+            "Stable_Properties",
             "Depends", "Refined_Depends",
             "Predicate_Failure"
         )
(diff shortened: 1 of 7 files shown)


ada/ast.py
```diff
@@ -8478,13 +8479,65 @@ def xref_equation():
...
+
+        return identifiers.logic_all(
+            lambda i: Self.env_get(
+                env,
+                i.sym,
+                lookup=LK.recursive,
+                from_node=Self.origin_node,
+                categories=all_categories
+            ).filtermap(
+                # It can only refers to a SubpDecl
```
Roldak, line 8527: `refers` -> `refer`

Roldak: Also, `SubpDecl` does not include `ExprFunction`, is that OK according to the AI?

ada/ast.py
```diff
@@ -13879,6 +13930,52 @@ def env_elements_baseid():
             )
         ))
 
+    @langkit_property(return_type=Equation)
+    def denotes_the_property_function(subp_decl=T.SubpDecl.entity):
```
Roldak, line 13934: There is something weird: since the equation returned is either `LogicTrue()` or `LogicFalse()`, it seems to me like you don't even need an equation here: you can simply return a `Bool` and use to filter out subprogram candidates directly in `stable_properties_assoc_equation`.

Roldak (changes requested):

ada/ast.py
thvnx: hm, I don't see anything that would forbid their use. If GNAT supports them, I'll fix the PR accordingly.

ada/ast.py
```diff
@@ -13930,50 +13930,45 @@ def env_elements_baseid():
...
-            LogicTrue(),
-            LogicFalse()
+        return And(
+            # It only has one single parameter (mode in but not checked here)
+            subp_spec.params.length == 1,
+
+            # It matches the type for which the Stable_Properties is defined.
+            # There are two cases:
+            Entity.parent_basic_decl.match(
+                # Either the Stable_Propeties aspect is defined within a
```
Roldak, line 13951: Propeties -> Properties

ada/ast.py
```diff
@@ -13930,50 +13930,45 @@ def env_elements_baseid():
...
+                # TypeDecl.
+                lambda td=T.TypeDecl: Not(
+                        primitive_types.find(
+                            lambda t:
+                            t.name.matches(td.name.node)
```
Roldak, line 13956: Why a name comparison? Could you not compare `t` and `td` directly? (same below)

ada/ast.py
```diff
@@ -8523,18 +8523,18 @@ def stable_properties_assoc_equation():
...
-                lambda f: f.is_a(T.SubpDecl)
+            ).filter(
+                # It can only refer to a SubpDecl or an ExprFunction
+                lambda f: Or(
+                    f.cast(T.SubpDecl).then(
+                        lambda f: i.denotes_the_property_function(f.subp_spec)
+                    ),
+                    f.cast(T.ExprFunction).then(
+                        lambda f: i.denotes_the_property_function(f.subp_spec)
+                    )
```
Roldak, line 8534: We have a `subp_spec_or_null` property, so you could simplify this code to something like:
```
lambda f: f.is_a(SubpDecl, ExprFunction) & i.denotes_the_property_function(f.cast(BasicDecl).subp_spec_or_null))
```

Roldak (changes requested):

ada/ast.py
thvnx: I was just confused I guess, thanks for the suggestion!

ada/ast.py
thvnx: Indeed, much better that way!

Roldak (approved):

Roldak (approved):

PR #915, opened 2022-01-11
Minimal update to python 3

petacreepers23:
This made this code usable in python 3.

Before it crashed by not truncating on division, or by the use of .itervalues instead of .values.

This is to comply with the python 3 enhancement proposed in 2019.

Also, the link to the CLA is broken, in the contributing guidelines.
_The first time you want to contribute a patch on the AdaCore GitHub, we ask you to electronically approve our Individual Contributor Licence Agreement. CLA assistant will send you a link on your GitHub pull request to do so._

Cheers.

### DIFF
--- a/contrib/detect_copy_paste_sa.py
+++ b/contrib/detect_copy_paste_sa.py
@@ -50,9 +50,9 @@ def suffix_array(s, k=256, n=None):
         n = len(s)
         s += [0, 0, 0]
 
-    n0 = (n + 2) / 3
-    n1 = (n + 1) / 3
-    n2 = n / 3
+    n0 = int((n + 2) / 3)
+    n1 = int((n + 1) / 3)
+    n2 = int(n / 3)
     n02 = n0 + n2
 
     s12 = [0] * (n02 + 3)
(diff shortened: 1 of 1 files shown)


contrib/detect_copy_paste_sa.py
```diff
@@ -50,9 +50,9 @@ def suffix_array(s, k=256, n=None):
         n = len(s)
         s += [0, 0, 0]
 
-    n0 = (n + 2) / 3
-    n1 = (n + 1) / 3
-    n2 = n / 3
+    n0 = int((n + 2) / 3)
```
pmderodat, line 53: There’s an idiomatic way to do this in Python3: `(n + 2) // 3` i.e. true integer division directly. Could you use this pattern instead of the float division + reconversion to int?

pmderodat (changes requested):
Thank you for the update! I think this is ok after the suggested change.

contrib/detect_copy_paste_sa.py
petacreepers23: It seems reasonable, it is implemented in next commit :)

pmderodat: Thanks!

pmderodat (approved):

PR #926, opened 2022-01-20
Add a new property to query the kind of a CallExpr

thvnx:
Add a new property `CallExpr.kind` to query the kind of a `CallExpr`
(call, array slice, array index, or type conversion). It returns a
`CallExprKind` enum value.

TN: UC22-030

### DIFF
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -11940,6 +11940,21 @@ class TargetName(Name):
         )
 
 
+class CallExprKind(Enum):
+    """
+    Kind of CallExpr type.
+
+    - ``call`` is when the CallExpr is a procedure or function call.
+    - ``array_slice``, ``array_index`` is when the CallExpr is in fact an
+      array slice or an array subcomponent access expression, respectively.
+    - ``type_conversion`` is when the CallExpr is a type conversion.
+    """
+    call = EnumValue()
+    array_slice = EnumValue()
+    array_index = EnumValue()
+    type_conversion = EnumValue()
+
+
 class CallExpr(Name):
     """
     Represent a syntactic call expression.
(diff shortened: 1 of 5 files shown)


ada/ast.py
```diff
@@ -11912,6 +11912,23 @@ def xref_equation():
...
+    """
+    Kind of CallExpr type.
+
+    - ``NoCallExpr`` is the null value for this enum.
+    - ``Call`` is when the CallExpr is a procedure or function call.
+    - ``ArraySlice``, ``ArrayIndex`` is when the CallExpr is in fact an array
+      slice or an array subcomponent access expression, respectively.
+    - ``TypeConversion`` is when the CallExpr is a type conversion.
+    """
+    NoCallExpr = EnumValue(is_default=True)
```
Roldak, line 11925: Is it necessary to have the `NoCallExpr` value? Maybe we should raise an exception instead?

testsuite/tests/properties/call_expr_kind/kind.adb
```diff
@@ -0,0 +1,48 @@
...
+   end F1;
+
+   Y : Boolean;
+begin
+   P.T2 (V3).A := P.T2 (V2).A;
+   --% node.f_dest.f_prefix.p_kind
+   --% node.f_expr.f_prefix.p_kind
+
+   Q := Integer (2.5);
+   --% node.f_expr.p_is_constant
```
Roldak, line 38: Not sure this belongs here! :smile:

Roldak (changes requested):
More general comment: It is said in the original ticket that this work could allow removing some duplicated logic. Could you find any opportunities to do so?

thvnx:
> More general comment: It is said in the original ticket that this work could allow removing some duplicated logic. Could you find any opportunities to do so?

Actually, not really. I just simplified `Name.is_constant` property but I didn't find any other place where it would be beneficial to call `Name.kind`. I could move the code for `is_call`, or `is_array_slice` here and call `kind` where they are used, not sure it worth the effort though.

Roldak:
> I could move the code for is_call, or is_array_slice here and call kind where they are used, not sure it worth the effort though.

I don't think it's a good idea either. Let's leave it like that then, thanks for investigating!

Roldak (dismissed):

Roldak (approved):

PR #927, opened 2022-01-21
Consider ForLoopVarDecl when building ForLoopSpec xref equation

thvnx:
Not sure adding the ForLoopVarDecl into the ForLoopSpec equation worth it since it is already indirectly handled in the current equation. So, I just created this PR to discuss it and see if I should go further in this direction.


thvnx:

PR #928, opened 2022-01-21
New properties for associating formal parameters to actual or default expressions

thvnx:
Add the new properties ``Name.call_params``, ``GenericInstantiation.inst_params``, ``BaseAggregate.aggregate_params``, ``SubtypeIndication.subtype_constraints``, and ``EnumRepClause.params``. 

TN: T622-004
TN: S617-020
TN: V202-011

### DIFF
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -7737,6 +7737,17 @@ class AnonymousType(TypeExpr):
     custom_id_text = Property(Entity.type_decl.text)
 
 
+class ParamActual(Struct):
+    """
+    Data structure used by zip_with_params, Name.call_params,
+    GenericInstantiation.inst_params, BaseAggregate.aggregate_params, and
+    SubtypeIndication.subtype_constraints properties. Associates an expression
+    (the actual) to a formal param declaration (the parameter).
+    """
+    param = UserField(type=T.DefiningName.entity)
+    actual = UserField(type=T.Expr.entity)
+
+
 class SubtypeIndication(TypeExpr):
     """
     Reference to a type by name.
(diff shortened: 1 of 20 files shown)


ada/ast.py
```diff
@@ -9101,6 +9110,40 @@ def actual_expr_decls():
         )
     )
 
+    @langkit_property(public=True, return_type=ParamActual.array)
+    def inst_params():
```
thvnx, line 9114: here

ada/ast.py
```diff
@@ -12881,20 +12949,41 @@ class MultiDimArrayAssoc(AggregateAssoc):
...
 
+    @langkit_property(return_type=T.Expr.entity)
+    def actual_for_param_at(param=T.DefiningName.entity,
+                            pos=T.Int,
+                            default_expr=(T.Expr.entity, No(T.Expr.entity))):
+        """
+        Return the actual expression for ``param`` if any, ``default_expr``
+        otherwise.
+        """
+        up = Var(Entity.unpacked_params)
```
Roldak, line 13062: It feels very sub-optimal to rebuild the whole flattened list of `unpacked_params` each time this property is called, which means each time we treat one argument in the calling property. Could there be a way to avoid it? Maybe by passing it as parameter, or even memoizing it.

testsuite/tests/properties/call_params/pkg.adb
```diff
@@ -0,0 +1,71 @@
+package body Pkg is
```
Roldak, line 1: Is there a difference between `pkg.adb` and `callp.adb` in terms of what you're testing? (Just our of curiosity, in case I missed something)

Roldak (changes requested):

testsuite/tests/properties/call_params/dot.adb
```diff
@@ -0,0 +1,43 @@
...
+   P2.P (0);
+   --% node.f_call.p_call_params
+   P2.P;
+   --% node.f_call.p_call_params
+   X.PT (0);
+   --% node.f_call.p_call_params
+   X.PT (V => 0);
+   --% node.f_call.p_call_params
+   X.PT;
+   --% node.f_call.p_call_params
```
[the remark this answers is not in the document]
thvnx: Works as expected. However, I found a bug for the following case:

`X.PT (V => 0);`, where the first parameter of the dot call where wrongly updated with the value of `V`.

ada/ast.py
thvnx: I chose to memoize it!

thvnx:

PR #929, opened 2022-01-24
V117-001: Fix behavior of ``P_Base_Subp_Declarations``.

Roldak:


ada/ast.py
```diff
@@ -2352,8 +2352,11 @@ def base_subp_declarations():
         NOTE: for the moment this only works for tagged types. Remains to be
         seen if we need to extend it.
         """
+        # We don't want to take into account Entity's metadata, as the result
+        # of this property shouldn't depend upon how this node was retrieved.
+        entity_no_md = Var(Entity.without_md.cast(BasicDecl))
```
raph-amiard, line 2357: Not sure why you have a var since you only use this once inside a then, seems like it would be better inlined ?

raph-amiard (changes requested):

ada/ast.py
Roldak: done!

raph-amiard (approved):

PR #934, opened 2022-02-01
Fix ``SingleTaskTypeDecl.next_part_for_decl``

thvnx:
This change fixes a bug where ``next_part_for_decl``, when called on a
``SingleTaskTypeDecl`` returns None instead of the next part of its
``SingleTaskDecl`` parent.

TN: SB08-025

### DIFF
--- a/ada/ast.py
+++ b/ada/ast.py
@@ -5461,6 +5461,11 @@ class BaseTypeDecl(BasicDecl):
     def next_part():
         """
         Returns the next part for this type decl.
+
+        .. note:: Since this property returns a ``BaseTypeDecl``, it cannot be
+            used to retrieve the next part of ``TaskTypeDecl`` and
+            ``ProtectedTypeDecl`` nodes as their next part is actually a
+            ``Body``. Use ``BasicDecl.next_part_for_decl`` for those instead.
         """
 
         return Entity.match(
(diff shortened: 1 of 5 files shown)


ada/ast.py
```diff
@@ -5461,6 +5461,11 @@ def previous_part(go_to_incomplete=(Bool, True)):
     def next_part():
         """
         Returns the next part for this type decl.
+
+        .. note:: As next_part returns a BaseTypeDecl, a less precise version
+        of next_part: BasicDecl.next_part_for_decl should be used when the
+        next_part isn't a BaseTypeDecl but a body, such as ProtectedBody
+        or TaskBody.
```
Roldak, line 5468: Good idea to comment! But I think the wording is a bit weird: what do you think of something like:
```
Since this property returns a `BaseTypeDecl`, it cannot be used to retrieve
the next part of `TaskTypeDecl` and `ProtectedTypeDecl` nodes as their
next part is actually a `Body`. Use `BasicDecl.next_part_for_decl` for those
instead.
```

pmderodat: Also remember that this is using Sphinx syntax, so code exceprts should be surrounded with double backquotes. :-)

Roldak (changes requested):

ada/ast.py
```diff
@@ -5461,6 +5461,11 @@ def previous_part(go_to_incomplete=(Bool, True)):
     def next_part():
         """
         Returns the next part for this type decl.
+
+        .. note:: Since this property returns a ``BaseTypeDecl``, it cannot be
+        used to retrieve the next part of ``TaskTypeDecl`` and
+        ``ProtectedTypeDecl`` nodes as their next part is actually a ``Body``.
+        Use ``BasicDecl.next_part_for_decl`` for those instead.
```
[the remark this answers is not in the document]
thvnx: thanks, should be good now.

Roldak (approved):

PR #941, opened 2022-02-10
Improve completion for `SubtypeIndication`

thvnx:
Only return type declarations for type completion.

TN: UA05-032


thvnx:

